PR #67, opened 2013-07-16
URLs in a['href'] should be escaped for LaTeX

hperl:
An URL such as in a `a['href']` can contain URL escape codes, which all start with a `%`. The LaTeX-converter should escape all `%` to `\%` using `escape()`, or otherwise the unescaped `%` will be treated like a comment in LaTeX, corrupting the file.

This pull request fixes this by calling `escape()` on the URL.

### DIFF
--- a/lib/kramdown/converter/latex.rb
+++ b/lib/kramdown/converter/latex.rb
@@ -206,9 +206,9 @@ module Kramdown
       def convert_a(el, opts)
         url = el.attr['href']
         if url =~ /^#/
-          "\\hyperlink{#{url[1..-1]}}{#{inner(el, opts)}}"
+          "\\hyperlink{#{escape(url[1..-1])}}{#{inner(el, opts)}}"
         else
-          "\\href{#{url}}{#{inner(el, opts)}}"
+          "\\href{#{escape(url)}}{#{inner(el, opts)}}"
         end
       end
 


mkamensky:
I think this is actually wrong, escaping breaks for instance urls with a tilde in them. The hyperref manual says that " The special characters # and  ̃ do not need to be escaped in any way.". Unfortunately, it does not say anything about other characters...

gettalong:
@mkamensky Thanks for your comment - do you have any test cases for this?

mkamensky:
Yes, this comes from my real life example:

    echo '<https://www.math.bgu.ac.il/~amyekut/teaching/2016-17/der-cats-III/course_page.html>' | ruby -rkramdown -e 'puts Kramdown::Document.new(ARGF.read).to_latex'

It produces

    \href{https://www.math.bgu.ac.il/\ensuremath{\sim}amyekut/teaching/2016-17/der-cats-III/course\_page.html}{https://www.math.bgu.ac.il/\ensuremath{\sim}amyekut/teaching/2016-17/der-cats-III/course\_page.html}

which does not compile. Thanks!

gettalong:
@mkamensky Sorry for the long delay...

I have tested this now with #, % and ~: Seems to work fine without escaping. As you mentioned, the hyperref package doesn't say much about escaping but I found [this answer](https://tex.stackexchange.com/a/334205) on StackExchange.

Therefore I have changed the code to *just* escape the percent sign and nothing else. Seems to do what it should so far.

Will be in the next release!

mkamensky:
@gettalong Thanks!

Issue #64, opened 2013-06-22
Three back ticks with a specific language


xuhdev:
I see that

```
~~~~ ruby
puts 'I am ruby'
~~~~
```

is supported, what about supporting

``````
```ruby
puts 'I am ruby'
```
``````

? This is used in some other systems so we could make our markdown files more portable.

Thanks!

Issue #53, opened 2013-05-06
Support GitHub style ```lang ...``` codeblocks


postmodern:
I noticed that kramdown already supports [specifying the language of a codeblock](http://kramdown.rubyforge.org/syntax.html#language-of-code-blocks). Kramdown could also support the GitHub flavoured codeblock syntax:

``````
```ruby
puts 'hello'
```
``````

This would allow kramdown to be more compatible with markdown written for redcarpet; currently used by GitHub and [rubydoc.org](http://rubydoc.org/).

Issue #31, opened 2012-12-09
Kramdown in place of Redcarpet


trans:
I have been using Redcarpet b/c for my application is is important to support GitHub compatibility. However, Redcarpet does not appear to work for JRuby, so I am looking at switching to Kramdown. But I am concerned for lack of "```" style code blocks.

Is this something that could a least be optional for Kramdown?

Also, are there any other features that Kramdown _doesn't_ support that Redcarpet does? I looked for a comparison but did not come across one.

PR #68, opened 2013-07-22
Create a GFM (Github Flavored Markdown) parser

plexus:
As a first step, it supports "fenced code blocks" delimited by three or more
backticks.

I am aware that this has been discussed in #15, #31, #53, and #64, and that the general concensus is that using tildes is a more standard and portable approach.

However in #53 it is also hinted that a pull request for a separate parser for this specific variety of Markdown might be accepted. And since the popularity of GFM is not decreasing, I would also argue it's a very useful feature to have.

This PR adds a `GFM` parser, including tests and docs.

``` ruby
Kramdown::Document.new(text, :input => 'GFM')
```

I am testing the water with this PR, to see what the current sentiments are regarding GFM. The parser currently only implements a single extra feature, but if this is deemed acceptable then at least it creates a place to expand GFM conformance. I'll be happy to respond to remarks as to make this worthy of merging.


ePirat:
Is this still in kramdown? I can't seem to use three backticks for code highlighting using jekyll with kramdown and input set to GFM.

gettalong:
@ePirat Yes, GFM is still in the kramdown distribution. You need to tell Jekyll that kramdown should use the GFM parser, not the standard kramdown parser - please ask the Jekyll people about this, they can help you with this.

ePirat:
@gettalong I actually did this, I am using the GFM option:

```yaml
kramdown:
  input: GFM
  syntax_highlighter: rouge
```

gettalong:
@ePirat Then you still have to ask the Jekyll guys why it doesn't work because I don't use Jekyll myself.

ePirat:
@gettalong Thanks, just wanted to make sure it's not an issue of kramdown before I ask them.

PR #206, opened 2014-12-30
Ignore empty title attributes

### LINKED ISSUE #199: Ignore empty title attributes
An image with explicitly specified empty title attribute leads to incorrect Markdown:
```
$ kramdown -ihtml -okramdown <<<'<img src="x" alt="" title="" />'
![](x "")
```

As mentioned [here](http://kramdown.gettalong.org/syntax.html#inline-links) the title must not be empty in Markdown and passing the output back to kramdown leads to strange behavior:
```
$ kramdown <<<'![](x "")'
<p>![](x “”)</p>
```

Please ignore empty title attributes as it already happens for links:
```
$ kramdown -ihtml -okramdown <<<'<a href="x" title="">y</a>'
[y](x)
```

winniehell:
Fixes #199.

The fix is not complete though as some tests still fail. This is because conversion is not reversible for empty title attributes:
```
$ kramdown -ihtml -okramdown <<< '<p><img src="a" alt="b" title="" /></p>'
![b](a)
```
but
```
$ kramdown -ikramdown -ohtml <<< '![b](a)'
<p><img src="a" alt="b" /></p>
```
and therefore `assert_equal` fails.

Any suggestions? Should I avoid these tests?

### DIFF
--- a/lib/kramdown/converter/kramdown.rb
+++ b/lib/kramdown/converter/kramdown.rb
@@ -280,7 +280,7 @@ module Kramdown
                   end
           "[#{inner(el, opts)}][#{index}]"
         else
-          title = el.attr['title'].to_s.empty? ? '' : ' "' + el.attr['title'].gsub(/"/, "&quot;") + '"'
+          title = parse_title(el.attr['title'])
           "[#{inner(el, opts)}](#{el.attr['href']}#{title})"
         end
       end
@@ -290,7 +290,7 @@ module Kramdown
         if el.attr['src'].empty?
           "![#{alt_text}]()"
         else
-          title = (el.attr['title'] ? ' "' + el.attr['title'].gsub(/"/, "&quot;") + '"' : '')
+          title = parse_title(el.attr['title'])
           link = if el.attr['src'].count("()") > 0
                    "<#{el.attr['src']}>"
                  else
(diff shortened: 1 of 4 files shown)


gettalong:
Thanks for the pull request!

Before I can merge it:

* The `#convert_title` method should be renamed to `#prepare_title` or something similar because all `#convert_NAME` methods indicate the conversion of an kramdown element.

* If you want to test HTML input, please create a new file with the extension ".htmlinput". At first I didn't know why you added the tests because the normal conversion direction is from text to HTML.

Thank you!

winniehell:
Sorry, i don't get the htmlinput mechanism. :disappointed: 

The bugfix does not touch the html-to-html conversion but only the html-to-kramdown conversion. I do not understand how i would write the test such that the backwards conversion (kramdown-to-html) is skipped.

gettalong:
The test cases were written originally written to be in the direction text -> HTML. However, later on other directions were added, like text -> kramdown -> HTML, HTML -> kramdown -> HTML, ...

Normally, when testing the HTML -> kramdown direction, the `.text` file is compared against the converted `.html` file. However, sometimes it is explicitly wanted to only test the HTML -> kramdown direction and not the other way around (like it is in this pull request). To facilitate this, you can create a file with the extension `.htmlinput` and the output file with `.text`. As the test code sees that there is no `.html` associated, the kramdown -> HTML direction is not tested.

winniehell:
Seems like I finally fixed it... :sparkles: :rocket:

gettalong:
Thanks for the fix! I have squashed the commits and attributed the final one to you.

winniehell:
Thank you for merging! :+1:

PR #210, opened 2015-01-18
allow toc_levels option to be of type Range

jkraemer:
This is more or less cosmetics but I just stumbled across this - the docs are talking of 'range syntax' but passing an actual `Range` for the `:toc_levels` option didn't work.


gettalong:
Thanks, this is a good idea.

However, the values of the array would need to be converted to integers in case the range isn't an integer range (see the array case). Please update your pull request in this regard.

jkraemer:
updated!

gettalong:
Thank you - I have cherry-picked your commit!

PR #219, opened 2015-02-11
Add PlainText converter.

nybblr:
On more than one occasion, I needed a way to convert a Markdown document to plain text for use in meta tags, emails, etc.

This makes a lot of sense to me in Kramdown core, any reason it wouldn't be a good addition?

Right now, this has no specs which is :-1: I'd love to get some help writing the specs using the example-based system you have setup, would someone be able to point me in the right direction?

### DIFF
--- a/lib/kramdown/converter/plain_text.rb
+++ b/lib/kramdown/converter/plain_text.rb
@@ -1,0 +1,66 @@
+# -*- coding: utf-8 -*-
+#
+#--
+# Copyright (C) 2009-2014 Thomas Leitner <t_leitner@gmx.at>
+#
+# This file is part of kramdown which is licensed under the MIT.
+#++
+#
+
+module Kramdown
+
+  module Converter
+
+    # Converts tree to plain text. Removes all formatting and attributes.
+    # This converter can be used to generate clean text for use in meta tags,
+    # plain text emails, etc.
+    class PlainText < Base
+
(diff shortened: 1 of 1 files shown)


gettalong:
I don't really get what you want to achieve and why you don't just use the source text. Do you want to remove the kramdown markup characters?

nybblr:
@gettalong Correct. For example, I have a bio in markdown:

~~~markdown
**Jo Smith** is an _awesome_ guy who works at [Jo Smith Inc.](http://josmith.me) He writes emails that frequently include `code snippets`.

```ruby
def block_of_code
  puts "block"
end
~~~

And need to extract this:

~~~
Jo Smith is an awesome guy who works at Jo Smith Inc. He writes emails that frequently include code snippets.

    def block_of_code
      puts "block"
    end
~~~

I needed to reuse the content in plain text settings. One specific example is the meta description tag, which shows up in Google searches, but I used it in a few other places.

gettalong:
Thanks for the explanation.

One thing that is missing is the `autoload` statement in `lib/kramdown/converter.rb`.

Have you done any tests for this converter to ensure it works with all element types? Once the `autoload` statement is in, the kramdown test suite should include it for some basic tests.

gettalong:
Will reopen this pull request when it is worked on again.

PR #222, opened 2015-02-17
Add pygments syntax highlighter.

nybblr:
Adds support for syntax highlighting with Pygments. It's an incredibly popular (if not the no. 1) code highlighter, so it seems Kramdown support would be a suitable addition =)

A few things are not quite fleshed out, thoughts would be :+1:

- [ ] I have not included the `<div highlight>` part because everything gets wrapped in another `<div class="highlighter-pygments">` tag. The outer tag feels wrong, are there historical reasons the outer div needs to be used? For drop-in compatibility with blogging platforms like Jekyll/Middleman, the structure needs to just have the `<div class="highlight">` wrapper.
- [ ] This file has to require the REXML stuff because it isn't req'd in the Utils file where it is referenced. Any issues moving the statement there?
- [ ] The HTML Utils helpers would make sense usable without mixing in, e.g. `::Kramdown::Utils::Html.html_attributes` with an `extend self`
- [ ] Needs specs. I'd appreciate direction/examples on where to start.

### DIFF
--- a/lib/kramdown/converter.rb
+++ b/lib/kramdown/converter.rb
@@ -31,7 +31,7 @@ module Kramdown
 
     configurable(:syntax_highlighter)
 
-    ["Coderay", "Rouge"].each do |klass_name|
+    ["Coderay", "Rouge", "Pygments"].each do |klass_name|
       kn_down = klass_name.downcase.intern
       add_syntax_highlighter(kn_down) do |converter, text, lang, type, opts|
         require "kramdown/converter/syntax_highlighter/#{kn_down}"
--- a/lib/kramdown/converter/syntax_highlighter/pygments.rb
+++ b/lib/kramdown/converter/syntax_highlighter/pygments.rb
@@ -1,0 +1,37 @@
+require 'rexml/parsers/baseparser'
+
+module Kramdown::Converter::SyntaxHighlighter
+  module Pygments
+    begin
+      require 'pygments'
+
+      # Highlighting via pygments is available if this constant is +true+.
+      AVAILABLE = true
+    rescue LoadError
+      AVAILABLE = false  # :nodoc:
+    end
+
+    extend ::Kramdown::Utils::Html
+
+    DEFAULTS = {
+      :startinline => true,
+      :encoding => 'utf-8',
+      :nowrap => true
+    }.freeze
+
+    def self.call(converter, text, lang, type, _unused_opts)
+      opts = DEFAULTS.merge(converter.options[:syntax_highlighter_opts])
+      lexer = lang || opts[:default_lang]
+      return nil unless lexer
+
+      code = ::Pygments.highlight(text, :lexer => lexer, :options => opts)
+
+      attrs = {
+        "class" => "language-#{lexer.gsub('+', '-')}",
+        "data-lang" => lexer
+      }
+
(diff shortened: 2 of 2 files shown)


gettalong:
Are you aware that there is already a pygments-stylesheet compatible syntax highlighter available in kramdown called rouge? It should be faster than any other solution since it is written in pure Ruby.

What would the benefits of the pygments.rb gem be?

I have looked through the source and moved the needed REXML require lines where they belong - thanks for pointing this out!

nybblr:
I was not aware it was such a \*good\* drop-in, thanks for the note.

Personally, I still think it would be good to have converter support as some folks may rely on some of Pygments' niche features, and it's a pretty lightweight addition (doesn't introduce any new dependencies), but I'll defer to your judgement on that.

Either way, I would still like to address the other two items (thanks for addressing the REXML bit!)

- [ ] I have not included the <div highlight> part because everything gets wrapped in another `<div class="highlighter-pygments">` tag. The outer tag feels wrong, are there historical reasons the outer div needs to be used? For drop-in compatibility with blogging platforms like Jekyll/Middleman, the structure needs to just have the `<div class="highlight">` wrapper.
- [ ] The HTML Utils helpers would make sense usable without mixing in, e.g. `::Kramdown::Utils::Html.html_attributes` with an `extend self`

Thanks!

gettalong:
Actually, I have to say that I am a bit reluctant to add a dependency on a library that uses an executable to do its work. If it were an interface to a C-library, it would not be optimal (because it won't work on JRuby, for example) but acceptable.

However, since I implemented the syntax highlighting interface for exactly this reason and pygments *is* an often used highlighting engine, I will accept the pull request.

As for your questions:

* The outer tag is used for identifying the highlighter used. What is inside depends on the highlighter itself. If it would create a `<div class="highlight">...</div>` tag, then it should be included. Otherwise standard stylesheets wouldn't work (I assume).

    You will find that the rouge highlighter adds a `<pre class="highlight">...</pre>` that is also not omitted.

    Drop-in compatibility doesn't come into play here as long as the output of the pygments syntax highlighter is not changed. And for kramdown compatibility reasons, the outer `<div>` tag has to remain.

* The whole HTML Utils helper class is not suited to being used standalone. However, if you don't want to use `extend` in the highlighter you can add `module_function :html_attributes` to the HTML utils module. Then this method will also be available standalone.

* As for tests: Have a look at `test/testcases/block/06_codeblock/highlighting-rouge.html` and its `.text` and `.options` file. You need to create a similar one for pygments. Same for `test/testcases/span/03_codespan/highlighting-rouge.html`.

* Documentation: See `doc/syntax_highlighter/rouge.page` for an example, the file should be called like the syntax highlighter name. Please also update `doc/documentation.template` with a link to the new highlighter. And please also add a concise section in the documentation regarding the requirements needed for getting the syntax highlighter running (if this is necessary, I have never used pygments nor pygments.rb).

* You set the encoding to UTF-8, however, this is not always the case. You should use `converter.root.options[:encoding]` as this contains the used encoding on Ruby >= 1.9. If the encoding cannot be determined, the statement returns `nil`. Then it is probably best to set it to a sane default like UTF-8.

Thanks for your pull request!

gettalong:
Just to let you know: I will release a new version at the end of this week. So if this should be shipped, it has to be finished by then.

gettalong:
Will reopen this pull request when it is worked on again.

PR #238, opened 2015-04-03
Enabled loading of external images

raehalme:
I needed to transform some markdown files into PDF, and noticed that kramdown tries to load all images from local files at least when rendering the PDF. I'm no Ruby expert, but managed to hack together a trivial change where images starting with http(s) are loaded "as-is" from the external source. 

If you think it's good enough, please feel free to include it. And thanks for creating such a nice tool!


gettalong:
Thanks for the pull request. However, since this would potentially lead to exploits I won't include the change.

PR #240, opened 2015-04-05
 Add math engine based on MathJax-node that outputs to MathML 

tmthrgd:
This adds a new math engine based on [MathJax-node](https://github.com/mathjax/MathJax-node). MathJax-node runs the MathJax conversion in node.js. mathjaxnode.rb converts the TeX code to MathML in a node.js subprocess using the MathJax-node bundled tex2mml command.


lib/kramdown/converter/math_engine/mathjaxnode.rb
```diff
@@ -0,0 +1,43 @@
...
+    end
+
+    def self.call(converter, el, opts)
+      type = el.options[:category]
+
+      cmd = [ T2MPATH ]
+      cmd << "--inline" unless type == :block
+      cmd << "--semantics" if converter.options[:math_engine_opts][:semantics] == true
+      cmd << "--notexhints" if converter.options[:math_engine_opts][:texhints] == false
+      result = IO.popen(cmd << el.value).read.strip
```
gettalong: Better not use a pipe when a simple ``result = `#{cmd << el.value}`.strip`` would suffice since some systems might not support pipes.

lib/kramdown/converter/math_engine/mathjaxnode.rb
```diff
@@ -0,0 +1,43 @@
...
+
+module Kramdown::Converter::MathEngine
+
+  # Uses the MathJax-node library for converting math formulas to MathML.
+  module MathjaxNode
+
+    # MathjaxNode is available if this constant is +true+.
+    AVAILABLE = begin
+      npm = %x{npm --global --depth=1 list MathJax-node}
+      T2MPATH = File.join npm.lines.first.strip, "node_modules/MathJax-node/bin/tex2mml"
```
gettalong: Please use parentheses around arguments.

lib/kramdown/converter/math_engine/mathjaxnode.rb
tmthrgd: It's not quite that easy. The raw value of the TeX is passed in the command line. If you use `#{cmd << el.value}`, or similar, it becomes trivial to craft input that will execute arbitrary programs. Using `IO.popen` with an array as the first argument causes ruby bypasses the shell: ["If cmd is an Array of String, then it will be used as the subprocess’s argv bypassing a shell."](http://ruby-doc.org/core-2.2.1/IO.html#method-c-popen). Doing it this way prevents those exploits.

lib/kramdown/converter/math_engine/mathjaxnode.rb
```diff
@@ -0,0 +1,43 @@
...
+      /MathJax-node@/ === npm.lines.drop(1).join("\n")
+    rescue
+      false
+    end
+
+    def self.call(converter, el, opts)
+      type = el.options[:category]
+
+      cmd = [ T2MPATH ]
```
gettalong: Please no spaces inside square brackets.

lib/kramdown/converter/math_engine/mathjaxnode.rb
gettalong: Okay, thanks for the explanation. I have been using the `systemu` library in such cases which should work on any system. Just that I know: Would `IO.popen` work on Windows?

gettalong:
Thanks for this math engine!

As for the name: I haven't done anything with node.js yet but I gather that node.js does have a Markdown library, so the target audience would be Ruby developers. Would it be better to name this engine something like 'MathJax-MathML'?

As for testing: Do you know what to put into the `.travis.yml` file to get the needed binaries and libraries? Some more explanation on the documentation page would also help people who do not know about node.js. I.e. explaining that an external program called node.js has to be installed as well as npm (if it is not packaged with node.js) and then the MathJax-Node library. (This would also help me set up my machine correctly for testing the engine ;-))

lib/kramdown/converter/math_engine/mathjaxnode.rb
tmthrgd: It certainly should. It appears that `systemu` itself [calls `IO.popen`](https://github.com/ahoward/systemu/blob/master/lib/systemu.rb#L84) to execute commands on all systems, with the exception of JRuby which is using [`java.lang.Runtime.runtime.exec`](https://github.com/ahoward/systemu/blob/master/lib/systemu.rb#L288).

tmthrgd:
I named it to be consistent with the underlying library *but* what your saying does make sense. I'll look into the `.travis.yml` and installing node.js bit at some point. Node.js just seems like ruby or python to me, I just sort of expect it to be there in my terminal. I'll get onto that though and clarify things.

gettalong:
Thanks!

tmthrgd:
All the errors attributed to 'Problem using tidy' look something like this:

```
line 1 column 1 - Warning: missing <!DOCTYPE> declaration
line 1 column 1 - Warning: inserting implicit <body>
line 1 column 12 - Error: <math> is not recognized!
line 1 column 12 - Warning: discarding unexpected <math>
line 2 column 3 - Error: <mi> is not recognized!
line 2 column 3 - Warning: discarding unexpected <mi>
line 2 column 8 - Warning: discarding unexpected </mi>
line 3 column 3 - Error: <mo> is not recognized!
line 3 column 3 - Warning: discarding unexpected <mo>
line 3 column 25 - Warning: discarding unexpected </mo>
line 4 column 3 - Error: <mi> is not recognized!
line 4 column 3 - Warning: discarding unexpected <mi>
line 4 column 8 - Warning: discarding unexpected </mi>
line 5 column 3 - Error: <mo> is not recognized!
line 5 column 3 - Warning: discarding unexpected <mo>
line 5 column 25 - Warning: discarding unexpected </mo>
line 6 column 3 - Error: <mo> is not recognized!
```

When tidy is removed the tests pass on my machine but fail when tidy is installed.

tmthrgd:
Unfortunately `IO.popen` has a different, incompatible, signature on ruby 1.8.7 which is causing that particular build to fail. All others are succeeding. The two solutions I can come up with are either: removing support for ruby 1.8.7 (extended maintenance did end July 31, 2014), or selectively disable Mathjax-Node on ruby < 1.9.

gettalong:
You can just remove support for Ruby 1.8.7 from the engine since the engine isn't bound by backwards compatibility yet.

tmthrgd:
Hurrah! All tests are passing. It's your call on the name though.

gettalong:
Thanks for all your work! Regarding the name: I will let the information sink in some time and then decide.

This will be in the next release.

tmthrgd:
:+1:

gettalong:
@tmthrgd I'm just integrating your code for the next release. After some problems with nodejs/npm that I needed to fix the tests work fine. However, they take *ages* to run. Is node.js that slow? Or has this to do with MathJax-node?

Anyway, the additional tests take a whooping 20 seconds to execute whereas the *whole* test suite needed 35 seconds before... That's not optimal. Do you know of a way to speed up the MathJax-node tests?

gettalong:
I have merged your commits into one and pushed to master, ie. your changes are now live.

Regarding the slow tests: I will modify the test suite so that slow tests are not executed depending on an environment variable.

gettalong:
@tmthrgd New commits fail in Travis due to an error with `tex2mml`: Unknown argument: f(x) = a{x^3} + b{x^2} + cx + d.
See for example https://travis-ci.org/gettalong/kramdown/builds/109940440.

When I test locally everything works fine - do you know why this happens?

tmthrgd:
@gettalong I was able to recreate the error locally after running `sudo npm --global install MathJax-node`. While I have no idea what was causing the problem, it is fixed by instead using the `mathjax-node`.

PR #241, opened 2015-04-08
Fix fenced code blocks

skorks:
Triple backtick with language wasn't working, this fixes it e.g.:

     ```ruby
     def blah
     end
     ```

wasn't working, but:

     ~~~ruby
     def blah
     end
     ~~~

was working fine - a regex issue.


gettalong:
This is not an issue but intentional. Triple backticks work when using the GFM parser. If you want the best compatibility among different Markdown parsers you should not use backticks but tildes for fenced codeblocks.

Issue #244, opened 2015-04-18
Support "-" (dash) in name of code block lang


dg-ratiodata:
Hi @gettalong,

are you going to accept a PR which makes it possible to use a dash in the code block language?

    ~~~ mermaid-graph
    ~~~

Maybe it's just this [line](https://github.com/gettalong/kramdown/blob/master/lib/kramdown/parser/kramdown/codeblock.rb#L35) which needs to be changed to

```ruby
FENCED_CODEBLOCK_MATCH = /^((~){3,})\s*?(\w+-)?\s*?\n(.*?)^\1\2*\s*?\n/m
```

gettalong:
Yes, I would accept it. I still have to have a look at the code to determine if this would have a side effect but I don't think so.

The change should be backwards-compatible because valid fenced code blocks are not affected.

maxmeyer:
I opened a new PR using my private account.

gettalong:
@maxmeyer  Just one question before we go further: Wouldn't it be easier to just use an underscore instead of a dash in your custom language? Or is there another reason?

maxmeyer:
> Wouldn't it be easier to just use an underscore instead of a dash in your custom language? 

To make it short: It's a reason for style and maybe compatibility with concurrent frameworks.

I would like to integrate [mermaid](https://github.com/knsv/mermaid) into [middleman-presentation](https://github.com/fedux-org/middleman-presentation) via "code blocks":

    ~~~
    graph TD;
    A-->B;
    A-->C;
    B-->D;
    C-->D;
    ~~~

I'm using a [custom kramdown code block converter](https://github.com/fedux-org/middleman-presentation/blob/95d55cebe43f3a961771185ea692e906ae1ec62a/middleman-presentation-core/lib/middleman-presentation-core/syntax_highlighter.rb) already so I'm able to generate the HTML stuff in a way I need. 

Using code blocks raise a problem "for me": What if a user wants to highlight the mermaid code instead of rendering a graph? It looks like the guys over [here](https://github.com/hakimel/reveal.js/issues/1082) decided to use something like this to make it possible:

    ~~~ mermaid-graph
    graph TD;
    A-->B;
    A-->C;
    B-->D;
    C-->D;
    ~~~

Personally I like the `mermaid-graph`-thing because I use the dash as delimiter and the underscore where text belongs together but whitespace cannot be used: "20140101-this_is_a_text"

maxmeyer:
BTW:
[...]

PR #245, opened 2015-04-18
Support dash in name of language

dg-ratiodata:
PR for #244.

### DIFF
--- a/lib/kramdown/parser/kramdown/codeblock.rb
+++ b/lib/kramdown/parser/kramdown/codeblock.rb
@@ -32,7 +32,7 @@ module Kramdown
 
 
       FENCED_CODEBLOCK_START = /^~{3,}/
-      FENCED_CODEBLOCK_MATCH = /^((~){3,})\s*?(\w+)?\s*?\n(.*?)^\1\2*\s*?\n/m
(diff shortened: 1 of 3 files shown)


dg-ratiodata:
This is not finished yet. I was not sure what is needed to run the test suite. I will update the PR, when I understand your development environment.

lib/kramdown/parser/kramdown/codeblock.rb
```diff
@@ -32,7 +32,7 @@ def parse_codeblock
 
 
       FENCED_CODEBLOCK_START = /^~{3,}/
-      FENCED_CODEBLOCK_MATCH = /^((~){3,})\s*?(\w+)?\s*?\n(.*?)^\1\2*\s*?\n/m
+      FENCED_CODEBLOCK_MATCH = /^((~){3,})\s*?(\w+-)?\s*?\n(.*?)^\1\2*\s*?\n/m
```
gettalong: The change would need to be from `(\w+)` to `(\w[\w-]*)`.

maxmeyer: Sure. Must have been blind. Fixed that, but forgot to push. :smile: But there's some more code which needs to be changed. Somewhere "you" may split the html-class-name. Only parts of the same go to `class="lang"`

gettalong:
See #246 for the new PR.

PR #246, opened 2015-04-19
Language name with dash

maxmeyer:
@gettalong 

A new one for #244. This [line](https://github.com/gettalong/kramdown/compare/master...maxmeyer:feature/name-with-dash?expand=1#diff-a97debc2020b21fffbb128d52c6f6a73R121) might be worth to mention. Without the addition `latex` is not able to compile the document. If you've got a better idea where I should place the code, I'm open for suggestions.


gettalong:
* The regexp should be `\w[\w-]*` so that it does not start with a dash.
* Please remove the `tmp/` directory from `.gitignore`.
* You need to add the test to the `EXCLUDE_LATEX_FILES` list to avoid running the conversion instead of the added code.

maxmeyer:
Fixed. But now some coderay tests are failing. Any idea how to fix that or is there another `EXCLUDE_` I can use for this?

maxmeyer:
Ok. Now I understand your test-setup better. I added an options-file as. Should I add a test for coderay as well?

doc/syntax.page
```diff
@@ -423,6 +423,14 @@ of the code block to the end of the starting line:
...
     ~~~
 
+You can also use "-" in the language name:
+
+    ~~~ ruby-trunk
+    def what?
+      42
+    end
+    ~~~
+
```
gettalong: Please only add an example that uses a valid, real world language. Otherwise people might think that `ruby-trunk` is some kind of Ruby based language.

Personally, I would not add any example because we never explicitly state the syntax requirements for the language. Also if a user wants to highlight a programming language that contains a dash he will just use it in the fenced code blocks.

lib/kramdown/converter/syntax_highlighter/coderay.rb
```diff
@@ -27,7 +27,7 @@ def self.call(converter, text, lang, type, _unused_opts)
       if type == :span && lang
         ::CodeRay.scan(text, lang.to_sym).html(options(converter, :span)).chomp
       elsif type == :block && (lang || options(converter, :default_lang))
-        lang = (lang || options(converter, :default_lang)).to_sym
+        lang = (lang || options(converter, :default_lang)).to_s.gsub(/-/, '_').sub(/language_/, 'language-').to_sym
         ::CodeRay.scan(text, lang).html(options(converter, :block)).chomp << "\n"
```
gettalong: The first `gsub` I understand but why the second `sub`?

test/testcases/block/06_codeblock/with_lang_in_fenced_block_name_with_dash.text
```diff
@@ -0,0 +1,33 @@
+~~~ ruby-trunk
+def what?
+  42
+end
+~~~
+
```
gettalong: Why all the tests below? They don't make sense for this feature enhancement.

test/testcases/block/06_codeblock/with_lang_in_fenced_block_name_with_dash.options
```diff
@@ -0,0 +1,2 @@
+:enable_coderay: false
+
```
gettalong: I think there shouldn't be any need for this file anymore since you modify the language name for coderay to use underscores instead of dashes.

test/testcases/block/06_codeblock/with_lang_in_fenced_block_name_with_dash.text
maxmeyer: To make sure, that a language with a dash behaves the same like a language without a dash

lib/kramdown/converter/syntax_highlighter/coderay.rb
maxmeyer: Fixed.

test/testcases/block/06_codeblock/with_lang_in_fenced_block_name_with_dash.options
maxmeyer: I added this because the HTML generated without the coderay converter is much simpler and is the same like for the existing tests.

doc/syntax.page
maxmeyer: Fixed.

test/testcases/block/06_codeblock/with_lang_in_fenced_block_name_with_dash.text
maxmeyer: Removed.

gettalong:
Thanks for everything and your patience! The changes are now in master.

maxmeyer:
Wonderful. :smile: Thanks a lot. Any plans for a new release?

gettalong:
Probably end of this week.

maxmeyer:
:+1:

PR #248, opened 2015-04-24
Use U+21A9 U+FE0E as the footnote backlink symbol (fixes #247)

### LINKED ISSUE #247: Allow the footnote backlink glyph to be changed
Kramdown uses “↩” (U+21A9) to link from footnotes back to the main text. Under OS X 10.10.3 and iOS 8.3, though, this character appears in Emoji form (↩️, if your browser supports it). [Like John Gruber](https://twitter.com/charlietuna/status/590362985861812225), I’d like to replace the bare U+21A9 with the two-character sequence U+21A9 U+FE0E, which will [force it to render in “text style”](http://mts.io/2015/04/21/unicode-symbol-render-text-emoji/). (That’s what I did in the the first sentence of this paragraph.)

Would it be possible to add a configuration option so that kramdown uses `&#x21A9;&#xFE0E;` as its footnote backlink “character”? Alternately, it might make sense just to make this sequence the default; I’d guess that more people prefer the old style symbol to the emoji symbol.

*(This feature request is copied from [this Stack Overflow question](http://stackoverflow.com/questions/29830488/is-it-possible-to-change-kramdown-s-use-of-↩-for-footnote-backlinks/29837706#29837706).)*

bdesham:


gettalong:
See explanation in #247.

PR #260, opened 2015-05-31
Prevent conversion to Emoji on iOS

leo:
Appending `&#xfe0e;` to the Unicode-represenation of the reversed arrow will prevent the conversion to an Emoji graphic on iOS. Take a look at this issue for better understanding: https://github.com/jekyll/jekyll/issues/3751

### DIFF
--- a/lib/kramdown/converter/html.rb
+++ b/lib/kramdown/converter/html.rb
@@ -443,9 +443,9 @@ module Kramdown
             insert_space = false
           end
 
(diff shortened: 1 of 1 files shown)


leo:
Sorry for not testing the change, I didn't know this was necessary. However, this probably won't affect anything, it's just a better representation of an unicode character. :innocent:

leo:
Here's the new one: https://github.com/gettalong/kramdown/pull/261

PR #261, opened 2015-05-31
Prevent conversion to Emoji on iOS

leo:
Appending `&#xfe0e;` to the Unicode-represenation of the reversed arrow will prevent the conversion to an Emoji graphic on iOS. Take a look at this issue for better understanding: https://github.com/jekyll/jekyll/issues/3751

### DIFF
--- a/lib/kramdown/converter/html.rb
+++ b/lib/kramdown/converter/html.rb
@@ -443,9 +443,9 @@ module Kramdown
             insert_space = false
           end
 
-          para.children << Element.new(:raw, FOOTNOTE_BACKLINK_FMT % [insert_space ? ' ' : '', name, "&#8617;"])
+          para.children << Element.new(:raw, FOOTNOTE_BACKLINK_FMT % [insert_space ? ' ' : '', name, "&#8617;&#xfe0e;"])
           (1..repeat).each do |index|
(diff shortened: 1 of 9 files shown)


gettalong:
See issue #247

leo:
Thank you for pointing me to this issue, @gettalong. However, I think the solution you provided is not fitting for me. You said that we could just post-process the output by using the following ruby code:

```Kramdown::Document.new(text).to_html.gsub(/\u21a9/, "\u21a9\ufe0e")```

The problem is, that I'm not able to use this code. The reason behind that is that I'm using Jekyll based on a Github pages site - we don't have the possibility to use our own ruby code.

You also said that 'Apple introduced this bug' and therefore should fix it. And I don't think it's a bug. It's just a behavior that's not necessarily bad - but we'd like to prevent it anyway, because it doesn't look that good on mobile devices.

Or do you see another way to replace this character on GitHub pages sites other than just merging this pull request?

gettalong:
If you use Github pages, you are bound by their limitations. That's part of the deal. If you don't want that, just generate the site locally where you have more options and push the result to Github. It's not that much more to do.

The way to change the default HTML output of kramdown is to subclass the HTML converter and change what you need. kramdown is a library and used by many libraries and applications, it is not just for Github pages.

PR #273, opened 2015-08-22
Implements a JSON converter

hectorcorrea:
Notice that I named the converter "Jsonx" to side-step an issue in which the code runs into conflicts with the Ruby's native "to_json". I believe the issue is in the code to automatically registers the converter but I am not sure. Thoughts?

If I name the class "Json" first call to to_json works as expected in the code below, but the second one does not.

```
puts Kramdown::Document.new(markdown).to_json
# => { the proper json goes here }

puts Kramdown::Document.new(markdown).to_json
# => "a string with the internal object ID is displayed here"
```


hectorcorrea:
As it turns out the Json gem is not bundled on Ruby 1.8.7 and hence the error. Thoughts on how to go about this?

hectorcorrea:
Closed in favor of an AST converter. See https://github.com/gettalong/kramdown/pull/275

PR #275, opened 2015-08-26
Implements an Abstract Syntax Tree converter that returns

hectorcorrea:
a Ruby hash with the tree for the elements in the document.


gettalong:
You should probably also add the `Element#options` to the output as they contain additional information like the line numbers.

hectorcorrea:
Good idea. Done.

gettalong:
This change has already been merged, see fb2efe23875df5ee2bd665b5dd14129a45d2d17a.

PR #276, opened 2015-08-27
Added space to force travis to rebuild and make sure all builds (incl…

hectorcorrea:
…uding Ruby 1.8.7) pass


hectorcorrea:
This was just a test to make sure all test pass in all versions of Ruby with the code in master.

PR #274, opened 2015-08-27
GFM parser looses line numbers.

mtuchowski:
While using `hard_wrap` it doesn't set elements location as can be seen here: https://github.com/gettalong/kramdown/blob/master/lib/kramdown/parser/gfm.rb#L41-L43

### DIFF
--- a/lib/kramdown/parser/gfm.rb
+++ b/lib/kramdown/parser/gfm.rb
@@ -38,8 +38,10 @@ module Kramdown
             omit_trailing_br = (Kramdown::Element.category(element) == :block && element.children[-1] == child &&
                                 lines[-1].empty?)
             lines.each_with_index do |line, index|
-              children << Element.new(:text, (index > 0 ? "\n#{line}" : line))
-              children << Element.new(:br) if index < lines.size - 2 ||
+              new_element_options = { :location => child.options[:location] + index }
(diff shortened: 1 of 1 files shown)


lib/kramdown/parser/gfm.rb
```diff
@@ -38,8 +38,10 @@ def add_hard_line_breaks(element)
             omit_trailing_br = (Kramdown::Element.category(element) == :block && element.children[-1] == child &&
                                 lines[-1].empty?)
             lines.each_with_index do |line, index|
-              children << Element.new(:text, (index > 0 ? "\n#{line}" : line))
-              children << Element.new(:br) if index < lines.size - 2 ||
+              new_element_options = { :location => element.options[:location] + index }
```
mtuchowski: There is still a problem with this in that `element.options[:location]` is always 1...

While tested with simple examples as here:

```markdown
one
two
three
```

it works as index is doing it's job, using `'` in text breaks the whole code :cry:

mtuchowski:
@gettalong Is there any chance you could look at this PR any time soon?

gettalong:
I will see what I can do.

gettalong:
Thanks for the pull request - I have applied the changes and added a test.

PR #282, opened 2015-09-08
Unicode support for footnote markers #281

### LINKED ISSUE #281: Unicode support for footnote definition
Since I write using non-ASCII characters, I realized it is not allowed to use non-ASCII footnote definitions such as `[^پانویس]`. So, here is the request to add Unicode support to them.

Ruby supports Unicode [pretty well](http://ruby-doc.org/core-1.9.3/Regexp.html#class-Regexp-label-Character+Properties) so it is basically possible.

mehdisadeghi:
This PR fixes #281 by adding Unicode equivalent for `\w` which is `\p{Alnum}` to kramdown's parser.

I had some problems with testing, since on my machine *many tests already fail*. Below test are the ones who fail with this change:
    Testing ./testcases/block/12_extension/options3.text                                    FAILED
    Testing ./testcases/block/12_extension/options.text                                     FAILED
    Testing ./testcases/block/16_toc/toc_exclude.text                                       FAILED
    Testing ./testcases/span/extension/options.text                                         FAILED 

If there are any issues let me know in order to fix them.

### DIFF
--- a/lib/kramdown/parser/kramdown/extensions.rb
+++ b/lib/kramdown/parser/kramdown/extensions.rb
@@ -125,9 +125,9 @@ module Kramdown
       end
 
 
-      ALD_ID_CHARS = /[\w-]/
+      ALD_ID_CHARS = /[\p{Alnum}-]/
       ALD_ANY_CHARS = /\\\}|[^\}]/
-      ALD_ID_NAME = /\w#{ALD_ID_CHARS}*/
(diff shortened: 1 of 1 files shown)


gettalong:
ALD names should not be changed when changing footnote marker names so this may be the reason why tests are failing. I'm also unsure whether the proposed changes would work on Ruby 1.8.

Please also see my comment on issue #281 which is another reason I'm closing this pull request.

PR #299, opened 2016-01-09
Add SmartyPants processor

pathawks:
This adds a minimal SmartyPants processor. It allows HTML to passthrough untouched, escapes special characters, and turns quotes into smart quotes and converts typographic characters.

### DIFF
--- a/lib/kramdown/parser/smartypants.rb
+++ b/lib/kramdown/parser/smartypants.rb
@@ -1,0 +1,17 @@
+# -*- coding: utf-8 -*-
+
+require 'kramdown/parser'
+
+module Kramdown
+  module Parser
+    class SmartyPants < Kramdown::Parser::Kramdown
+
+      def initialize(source, options)
+        super
+        @block_parsers = [:block_html]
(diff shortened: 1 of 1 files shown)


gettalong:
I don't think that this is something that is very useful to have in kramdown itself.

pathawks:
Fair enough :beers:

jhabdas:
@gettalong I was looking over the kramdown documentation today and didn't see an option for automatically outputting angle/curly quotes, or, what many Medium bloggers have come to expect, automatic rendering of `--` as an ndash. Am I missing something?

pathawks:
@jhabdas
https://kramdown.gettalong.org/options.html#option-smart-quotes
https://kramdown.gettalong.org/syntax.html#typographic-symbols

jhabdas:
Oof. Sorry I missed that. When I visited the site earlier I hit the options page @pathawks linked and did a full-text page search for "dash" and "hyphen" and came up empty-handed. I didn't even notice the other page. I'll do some more digging to figure out why my Jekyll theme isn't handling the typographic symbols as expected (what I was really after to begin with).

PR #184, opened 2014-11-16
Added strikethrough syntax to GitHub Flavored Markdown parser

diegobg:
GitHub Flavored Markdown supports strikethrough text according to [GitHub Help - GitHub Flavored Markdown](https://help.github.com/articles/github-flavored-markdown/) but the GFM parser included with kramdown doesn't support it.

I've added strikethrough support to the GFM parser: when it encounters text delimited by two or more tildes, it creates a &lt;del&gt; HTML element that contains the text. This doesn't affect the standard kramdown parser.

PR #307, opened 2016-02-08
Added strikethrough syntax to GitHub Flavored Markdown parser

### LINKED ISSUE #304: strikethrough doesn't work for GFM parser
GitHub Flavored Markdown supports strikethrough but the GFM parser included with kramdown doesn't support it.

In a [sample site](https://github.com/CGarces/strikethrough_issue) with the default github values

```
 kramdown:
   input: GFM
   hard_wrap: false
```

The strikethrough not work [see rendered page](http://cgarces.github.io/strikethrough_issue/)

Now is posible to made the strikethrough works using redcarpet.

```
markdown: redcarpet
redcarpet:
  extensions: ["strikethrough"]
```
But it'll be not possible in the future due the github pages migration to kramdown.

This is related with the PR #184

parkr:
Duplicates https://github.com/gettalong/kramdown/pull/184 by @diegobg. Fixes #304.

@gettalong We'd like to work with you to bring this GFM parser up to speed with the current offerings on GitHub. Expect some PR's to that effect in the coming weeks. Thank you for kramdown!

### DIFF
--- a/doc/parser/gfm.page
+++ b/doc/parser/gfm.page
@@ -19,6 +19,7 @@ changes:
 * Support for fenced code blocks using three or more backticks has been added.
 * Hard line breaks in paragraphs are enforced by default (see option 'hard_wrap').
 * ATX headers need a whitespace character after the hash signs.
+* Strikethroughs can be created using two or more tildes surrounding a piece of text
 
 ## Options
 
--- a/doc/parser/gfm.page
+++ b/doc/parser/gfm.page
@@ -19,7 +19,7 @@ changes:
 * Support for fenced code blocks using three or more backticks has been added.
 * Hard line breaks in paragraphs are enforced by default (see option 'hard_wrap').
 * ATX headers need a whitespace character after the hash signs.
-* Strikethroughs can be created using two or more tildes surrounding a piece of text
+* Strikethroughs can be created using two tildes surrounding a piece of text
 
 ## Options
 
--- a/lib/kramdown/parser/gfm.rb
+++ b/lib/kramdown/parser/gfm.rb
@@ -23,6 +23,10 @@ module Kramdown
           @block_parsers.delete(current)
           @block_parsers.insert(i, replacement)
         end
+
+        i = @span_parsers.index(:escaped_chars)
+        @span_parsers[i] = :escaped_chars_gfm if i
+        @span_parsers << :strikethrough_gfm
       end
 
       def parse
@@ -60,6 +64,30 @@ module Kramdown
       FENCED_CODEBLOCK_MATCH = /^(([~`]){3,})\s*?((\w[\w-]*)(?:\?\S*)?)?\s*?\n(.*?)^\1\2*\s*?\n/m
       define_parser(:codeblock_fenced_gfm, /^[~`]{3,}/, nil, 'parse_codeblock_fenced')
 
+      STRIKETHROUGH_DELIM = /~{2,}/
+      STRIKETHROUGH_MATCH = /#{STRIKETHROUGH_DELIM}.+#{STRIKETHROUGH_DELIM}/m
+      define_parser(:strikethrough_gfm, STRIKETHROUGH_MATCH, '~~')
+
+      def parse_strikethrough_gfm
+        line_number = @src.current_line_number
+
+        if @src.scan(STRIKETHROUGH_DELIM)
+          el = Element.new(:html_element, 'del', {}, :category => :span, :line => line_number)
+          @tree.children << el
+          parse_spans(el, STRIKETHROUGH_DELIM)
+          @src.scan(STRIKETHROUGH_DELIM)
+        end
+      end
+
+      ESCAPED_CHARS_GFM = /\\([\\.*_+`<>()\[\]{}#!:\|"'\$=\-~])/
+
+      # Parse the backslash-escaped character at the current location.
+      def parse_escaped_chars_gfm
+        @src.pos += @src.matched_size
+        add_text(@src[1])
+      end
+
+      define_parser(:escaped_chars_gfm, ESCAPED_CHARS_GFM, '\\\\')
     end
   end
 end
--- a/lib/kramdown/parser/gfm.rb
+++ b/lib/kramdown/parser/gfm.rb
@@ -64,8 +64,8 @@ module Kramdown
       FENCED_CODEBLOCK_MATCH = /^(([~`]){3,})\s*?((\w[\w-]*)(?:\?\S*)?)?\s*?\n(.*?)^\1\2*\s*?\n/m
       define_parser(:codeblock_fenced_gfm, /^[~`]{3,}/, nil, 'parse_codeblock_fenced')
 
-      STRIKETHROUGH_DELIM = /~{2,}/
-      STRIKETHROUGH_MATCH = /#{STRIKETHROUGH_DELIM}.+#{STRIKETHROUGH_DELIM}/m
+      STRIKETHROUGH_DELIM = /~~/
+      STRIKETHROUGH_MATCH = /#{STRIKETHROUGH_DELIM}[^\s~](.*)[^\s~]#{STRIKETHROUGH_DELIM}/m
       define_parser(:strikethrough_gfm, STRIKETHROUGH_MATCH, '~~')
 
       def parse_strikethrough_gfm
--- a/test/test_files.rb
+++ b/test/test_files.rb
@@ -299,6 +299,7 @@ class TestFiles < Minitest::Test
   # Generate test methods for gfm-to-html conversion
   Dir[File.dirname(__FILE__) + '/{testcases,testcases_gfm}/**/*.text'].each do |text_file|
     next if EXCLUDE_GFM_FILES.any? {|f| text_file =~ /#{f}$/}
+    next if RUBY_VERSION < '1.9'
     basename = text_file.sub(/\.text$/, '')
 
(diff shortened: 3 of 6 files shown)


parkr:
@gettalong Do you require a certain Ubuntu version for Travis?

```text
E: Unable to locate package texlive-latex-base
E: Unable to locate package texlive-latex-recommended
E: Unable to locate package texlive-latex-extra
```

parkr:
@gettalong Tests are passing on all but Ruby 1.8.7. May I ask why you still support 1.8.7? It's been EOL'd a really long time ago and it's generally A Bad Idea to use it now after a number of years in disrepair. [The 1.8.7 error](https://travis-ci.org/gettalong/kramdown/jobs/107875585#L579-L580):

```diff
-<p>I <del>don\342\200\231t even</del>~ have an extra tilde.</p>
+<p>I <del>don&#8217;t even</del>~ have an extra tilde.</p>
```

parkr:
@gettalong Do you have time to review this? If not, would you consider adding me as a collaborator?

geraldb:
@parkr  FYI: I have no inside knowledge (but I have confirmed Thomas Leitner for a talk on Tue Feb/23 in Vienna) - as far as I know Thomas might be on skiing vacations - thus, you will need a little patience until you will get a response.

parkr:
@geraldb Good to know! Thanks for the update.

gettalong:
@parkr As @geraldb mentioned I was on vacation the last fortnight - I will try to look at the issues/pull requests the next few days.

doc/parser/gfm.page
```diff
@@ -19,6 +19,7 @@ changes:
 * Support for fenced code blocks using three or more backticks has been added.
 * Hard line breaks in paragraphs are enforced by default (see option 'hard_wrap').
 * ATX headers need a whitespace character after the hash signs.
+* Strikethroughs can be created using two or more tildes surrounding a piece of text
```
gettalong: It seems that only two tildes are only ever used, not "two or more".

lib/kramdown/parser/gfm.rb
```diff
@@ -60,6 +64,30 @@ def add_hard_line_breaks(element)
       FENCED_CODEBLOCK_MATCH = /^(([~`]){3,})\s*?(\w[\w-]*)?\s*?\n(.*?)^\1\2*\s*?\n/m
       define_parser(:codeblock_fenced_gfm, /^[~`]{3,}/, nil, 'parse_codeblock_fenced')
 
+      STRIKETHROUGH_DELIM = /~~/
+      STRIKETHROUGH_MATCH = /#{STRIKETHROUGH_DELIM}[^\s~](.*)[^\s~]#{STRIKETHROUGH_DELIM}/m
+      define_parser(:strikethrough_gfm, STRIKETHROUGH_MATCH, '~~')
```
gettalong: Why are you potentially matching more than needed? It should be sufficient to use `STRIKETHROUGH_DELIM` here instead of `STRIKETHROUGH_MATCH`.

gettalong:
@parkr You can ignore the 1.8.7 failure for GFM, this is restriction is only for the main kramdown parser. As why it is still supported: There is currently no reason to not support it in 1.x and there *are* apparently folks still using 1.8.x.

parkr:
@gettalong Updated. I also skipped all GFM tests for 1.8.7.

gettalong:
Thanks! One more thing: Please combine all GFM strikethrough related commits (ie. all except the one concerning TravisCI) into one commit.

gettalong:
@parkr After the STRIKETHROUGH_DELIM change the test is failing because whitespace after the initial/before the final delimiter is not correctly handled.

Two options: Either restore the original behavior (which is potentially much slower) or look at the emphasis parsing implementation to see how to handle this situation.

parkr:
> look at the emphasis parsing implementation to see how to handle this situation.

Are you suggesting that I copy the implementation/logic for the emphasis parser?

gettalong:
You probably don't need all the logic from there but some logic needs to be included to make sure

* that the final `~~` is really found and not just the end of the string
* that there is no whitespace after the initial and before the final delimiter

gettalong:
@parkr The mathjax-node related errors are now fixed thanks to @tmthrgd. I have also already incorporated your Travis fix into the repo.

parkr:
@gettalong Rebased on latest `master`.

> that the final `~~` is really found and not just the end of the string

Added test case. It passes.

> that there is no whitespace after the initial and before the final delimiter

Added test case. It passes.

Anything else? I don't know a better way to accomplish this that's more efficient. Feel free to take our two commits and improve upon them, but I believe this works just fine. :smile:

parkr:
@gettalong Let me know if there is anything else needed here. I would love to see this feature make it into the next version of kramdown. Thanks!

gettalong:
There is still at least one edge case that is not handled correctly with the current code - I will look at it.

gettalong:
Before I do anything further could someone please tell me the GFM specification for strikethrough?

There are two possibilities:

1. Test any input in an input field like this one in which I'm writing and *try* to do whatever we can find out.

2. Do the correct thing, even for edge cases.

The reason why I ask is the following example:

~~~ markdown
This ~~is a complex strike through *test ~~with nesting~~ involved*.
~~~

This ~~is a complex strike through *test ~~with nesting~~ involved* .

It currently gives the following result when evaluated in this input field:

~~~ html 
<p>This <del>is a complex strike through *test ~~with nesting</del> involved*.</p>
~~~

As you can see this is not good...

So, what should be done? Support GFM like it is now or support something working in a useful way?

parkr:
Thanks for looking into that.

> So, what should be done? Support GFM like it is now or support something working in a useful way?

This is a question for @benbalter. My thoughts are to support it as it is rather than how it ought to be. If it makes our code _more_ complex to support it as it is, then I am fine with this breakage. :+1:

gettalong:
Since GFM is not my main concern it would be okay if it is broken by design (in contrast to the kramdown parser). However, note that if this is how the GFM parser should work it will be announced as such in kramdown's documentation to make this clear and tests for the broken design will also be added.

I also had to make this choice for kramdown initially: Support the broken Markdown.pl way or break 100% compatibility to do something useful and, especially, predictive. The latter was my choice for kramdown (though there are some things I would change).

benbalter:
Until we move to something like CommonMark, there's no such thing as a GFM spec, AFAIK, so there's not necessarily a "right" way to do things, in my mind. 

> Since GFM is not my main concern it would be okay if it is broken by design (in contrast to the kramdown parser).

@gettalong based on your comment above, it sounds like you see the GFM parser primarily as something to give parity to GitHub Markdown. If that's the case, then my vote would be for mirroring GitHub's implementation, even if it's silly. 

My reasoning is that if I push a file to GitHub.com (which uses GFM), and then they try to render it using Kramdown (which advertises GFM support) and the two differ, my first reaction would be to file a bug, because the software violated my expectations (regardless of if the behavior was right or not). 

If you see GFM support as a first-class, but alternative parser for those that prefer GFM (but may not use it with regard to GitHub's implementation), then the more philosophically pure implementation makes sense. Otherwise, I'd err on the side of practicality and user expectations, unless that adds unnecessary complexity to the code.

FWIW, I wrote up a quick test to compare common markdown fixtures rendered with Kramdown and with the GitHub Markdown API, and these seems to be the only major point of differentiation, once e.g., whitespace is normalized.

(Also to note, I did some spelunking through the github-markdown implementation and history, but couldn't find any guidance one way or the other).

geraldb:
If I may add my two cents @gettalong @benbalter @parkr

Why not split-off the kramdown GFM Parser class into a new gem and its own project e.g. kramdown-gfm or something. Why? First the code is actually so far minor e.g. about 100+ lines of ruby - see [Kramdown::Parser::GFM](https://github.com/gettalong/kramdown/blob/master/lib/kramdown/parser/gfm.rb). 

Thomas Leitner - as I talked to him - if I dare to say has no interest really in support the GitHub-Flavored (GFM) markdown (also as mentioned before - Thomas is not using Jekyll either ;-). His main concern is "core" kramdown. If you use your own (split-off) addon and gem than you can change the code as you like and that's in the best interest and you will get his support and approval of Thomas Leitner as far as I understand (from our in person conversations). And it make sense. As a real-world example there's already a [kramdon-rfc2629 parser](https://github.com/cabo/kramdown-rfc2629).

Why should Thomas in his spare time support the GitHub-Flavored (GFM) addon / syntax? I'd say that is not really optimal. Using your own kramdown-gfm addon - you can update and push changes whenver you like and Thomas and others can contribute (it's open source on GitHub ;-)  And again that code involved is minimal - maybe two or three hundred lines of Ruby?  Really not much more involved than a Jekyll plugin.  Than it's easy to add strikethrough, and some other minor missing extras - no need for a spec - it's your code and that's good enough (it's the spec ;-)

gettalong:
@benbalter I don't intend the GFM parser to be a standalone, separate Markdown parser that is used outside the context of Github and therefore "first-class". GFM makes, in my opinion, only sense when used in the context of Github, eg. Github Pages.

So if there is no GFM spec to work with or test against, I'm fine with including the changes in this way.

@geraldb Splitting off the GFM parser would be a possibility but is currently not really necessary as contributing to the GFM parser inside the kramdown project is possible.

gettalong:
The parser needed some more work to mimic Github in respect to my "broken" example. I also removed some unneeded code in regards to the escaping mechanism and fixed the tests.

Thanks for the input and discussion!

parkr:
@gettalong Thank you! And thank you for preserving our commits in your merge. Cheers! :beers:

PR #312, opened 2016-02-17
Fix bug that "input" cannot be recognized as an option

yzyzsun:
I found symbol `:input` can be recognized as an option of `Kramdown::Document.new`, but string `"input"` cannot. Then I found out that it is because `@` is omitted in `initialize` method.

This can also fix that `input` option does not work in [Jekyll](https://github.com/jekyll/jekyll)'s  `_config.yml` since Jekyll passes `input` as a string.


gettalong:
Thanks for the bug fix!

PR #313, opened 2016-02-18
Update MathJax-node to mathjax-node (note case)

tmthrgd:
Fix for failure reported in #240.

Replaces the deprecated `MathJax-node` library with `mathjax-node` (note the difference in case).


tmthrgd:
@gettalong (#240) this now seems to have caused a new issue with the version of Node.js.

tmthrgd:
@gettalong the failure reported in #240 is fixed in dfa8a26.

tmthrgd:
N.B.: For reference, the line to install nodejs is from [Austin Pray](http://austinpray.com/ops/2015/09/20/change-travis-node-version.html).

gettalong:
@tmthrgd Thanks for the quick response!

I have squashed your commits into one and additionally redirected stderr when checking for the library to avoid unnecessary output, see e9dbe88 .

PR #341, opened 2016-05-26
List supported parsers in Document docs

getaaron:
It took me longer than it should to figure out how to get from `Kramdown::Document.new(text)` to `Kramdown::Document.new(text, :input => 'GFM')`. I couldn't find "GFM" in the docs (I had to read the source code to find it).

I suggest listing the available options at the same point where the `:input` param is mentioned. This PR does just that.

One downside of this approach is that it's an extra maintenance point which will require updating when a new parser is added. I think the added clarity for developers is worth the extra maintenance.

### DIFF
--- a/lib/kramdown/document.rb
+++ b/lib/kramdown/document.rb
@@ -92,9 +92,8 @@ module Kramdown
     # options that can be used are defined in the Options module.
     #
     # The special options key :input can be used to select the parser that should parse the
-    # +source+. It has to be the name of a class in the Kramdown::Parser module. For example, to
-    # select the kramdown parser, one would set the :input key to +Kramdown+. If this key is not
-    # set, it defaults to +Kramdown+.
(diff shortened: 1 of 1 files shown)


gettalong:
If you read the API documentation at http://kramdown.gettalong.org/rdoc/Kramdown/Document.html, you will find that it says: "The special options key :input can be used to select the parser that should parse the source. It has to be the name of a class in the Kramdown::Parser module."

And when you view the availabel classes on the left side the API documentation (e.g. at http://kramdown.gettalong.org/rdoc/), you can easily see which parsers are included in the kramdown distribution.

So everything is already in the docs. And as you mention: Duplicating that information might not be good from a maintenance perspective.

getaaron:
Well yeah, that's what I did. I'm just saying I found that process to be an annoying amount of research required for a primary feature of this software. Perhaps the first statement could at least link to the appropriate place in the rdoc?

PR #345, opened 2016-06-08
change Config::CONFIG to RbConfig::CONFIG

boutil:
Config has been deprecated for a long time in favor of RbConfig and does not exist anymore since ruby2.2
In Debian, we rely on the 'datadir' config option to find the installed data.


gettalong:
Thanks for your patch, will be in the next release.

PR #356, opened 2016-07-21
No need to remove formatter from options

le0pard:
If you remove formatter for options, this leave next code without this formatter. Example without fix:

no fix

With fix:

fix

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: ruby
 rvm:
-  - 1.8.7
   - 1.9.3
   - 2.0.0
   - 2.1.0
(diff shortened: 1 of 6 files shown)


le0pard:
Looks like rouge 2.0 broke tests

gettalong:
Please provide the kramdown sources and their HTML output for the failing case, images are really useful.

le0pard:
I tried as I can, but for ruby 1.8 need more (not sure, why it supported) works. Also test env unstable (`npm install` network issue, rouge gem version 1 and 2)

le0pard:
@gettalong should I do something else?

gettalong:
Please provide the kramdown sources and their HTML output for the failing case here in the issue , images aren't really useful.

le0pard:
I just add this as test case. Or it is not enough?

gettalong:
kramdown test cases are plain text files with accompanying HTML and/or other output files, see test/testcases/.

It would be easier if you just provide the plain text source as well as the current HTML output and the wanted HTML output directly here so that one doesn't have to extract them from the code.

le0pard:
@gettalong looks like done. Please, check.

gettalong:
Thanks, I get the problem now (it's really too hot in the city to think straight...)!

gettalong:
I have updated the `test_files.rb` script to work with Ruby 1.8.7, locally works fine.

Change will be in the next release.

le0pard:
@gettalong thanks!

PR #357, opened 2016-07-27
Strip transliterated head ids

tsl0922:
source:

```
## Transliterated: 标题二
```

result:

```
<h2 id="biao-ti-er-">标题二</h2>
```

There's an extra `-` at the end of the header id, which is replaced from whitespace, we should strip it.

### DIFF
--- a/lib/kramdown/converter/base.rb
+++ b/lib/kramdown/converter/base.rb
@@ -230,7 +230,7 @@ module Kramdown
       # Uses the option +auto_id_prefix+: the value of this option is prepended to every generated
       # ID.
       def generate_id(str)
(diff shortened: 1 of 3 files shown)


gettalong:
Thanks but since this is not a bug, just a feature enhancement, this would change the output for people using the transliterated_header_ids option and therefore can't be done for the current 1.x series.

gettalong:
@tsl0922 Don't worry about CI failures right now, there will be more since this will only be added for the 2.x version of kramdown and there will be other commits in the master branch before 2.x. is started.

gettalong:
Closed because pull requests for features are now tracked through a dedicated project - see https://github.com/gettalong/kramdown/projects/1

PR #369, opened 2016-09-06
GFM: be more clever on parsing fenced codeblock

tsl0922:
This pr adds support for fenced code block with whitespace before the ```

source:

``````
  ```js
  console.log("hello");
  ```
``````

github result:

``` js
  console.log("hello");
```

kramdown result:

```
<p><code>js
  console.log("hello");
 </code></p>
```

### DIFF
--- a/lib/kramdown/parser/gfm.rb
+++ b/lib/kramdown/parser/gfm.rb
@@ -85,8 +85,8 @@ module Kramdown
         true
       end
 
-      FENCED_CODEBLOCK_START = /^[~`]{3,}/
-      FENCED_CODEBLOCK_MATCH = /^(([~`]){3,})\s*?((\S+?)(?:\?\S*)?)?\s*?\n(.*?)^\1\2*\s*?\n/m
+      FENCED_CODEBLOCK_START = /^[ ]{0,3}[~`]{3,}/
+      FENCED_CODEBLOCK_MATCH = /^[ ]{0,3}(([~`]){3,})\s*?((\S+?)(?:\?\S*)?)?\s*?\n(.*?)^[ ]{0,3}\1\2*\s*?\n/m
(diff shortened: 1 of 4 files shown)


lib/kramdown/parser/gfm.rb
```diff
@@ -85,8 +85,8 @@ def parse_atx_header_gfm_quirk
         true
       end
 
-      FENCED_CODEBLOCK_START = /^[~`]{3,}/
-      FENCED_CODEBLOCK_MATCH = /^(([~`]){3,})\s*?((\S+?)(?:\?\S*)?)?\s*?\n(.*?)^\1\2*\s*?\n/m
+      FENCED_CODEBLOCK_START = /^\s*?[~`]{3,}/
```
gettalong: `\s*?` is not correct since after testing I found that only a maximum of three spaces is allowed by GFM on Github.

lib/kramdown/parser/gfm.rb
```diff
@@ -85,8 +85,8 @@ def parse_atx_header_gfm_quirk
...
+      FENCED_CODEBLOCK_MATCH = /^\s*?(([~`]){3,})\s*?((\S+?)(?:\?\S*)?)?\s*?\n(.*?)^\s*?\1\2*\s*?\n/m
```
gettalong: Here is the same problem as in the last line.

lib/kramdown/parser/gfm.rb
tsl0922: Yes, it can be more strictly here, changed to `[ ]{,3}`. More than 3 spaces will be identified as normal code block at github, so does kramdown.

tsl0922: `[ ]{,3}` failed travis test for 1.8.7, changed it to `[ ]{0,3}`

gettalong:
Thanks for the changes - this looks good to me now! I will include it for the next release.

gettalong:
Thanks for the pull request - will be in the next release.

PR #372, opened 2016-09-30
Add preview_as_code option to mathjax engine

qwtel:
Adds a `preview_as_code` option to the mathjax engine.

When `true`,  wraps LaTeX previews in `code` / `pre > code` tags instead of `span` / `div` tags, mirroring the structure of inline code / code blocks. This ensures the content is styled like regular code and improves legibility in RSS readers like feedly or text browsers like lynx, etc. 

See in action:
Preview latex code


gettalong:
@qwtel Thanks for the pull request - this will be in the next release!

PR #378, opened 2016-10-15
Exclude mathjax tests when latex isn't present

agriffis:
This allows the test suite to run cleanly without latex.


agriffis:
Sorry, this was conflating two issues. The mathjax tests are broken right now, but it's not because of Latex's absence.

PR #379, opened 2016-10-16
Add :disable_parsers option.

agriffis:
I'm using kramdown with jekyll, and I'd like to defer smart quotes and typographic dashes to [rubypants](https://github.com/jmcnevin/rubypants) via [jekyll-pants](https://github.com/scampersand/jekyll-pants). However kramdown makes this hard/impossible because there's no way to disable the built-in parsers.

(There is a workaround for smart quotes, which is to set `smart_quotes: apos,apos,quot,quot`, but this is clearly a hack and doesn't help for dashes and ellipses.)

This pull request adds :disable_parsers so that one could set `disable_parsers: typographic_syms,smart_quotes` to properly defer that processing. I made it generic rather than specific to those parsers since it seems like it could be generally useful.

### DIFF
--- a/lib/kramdown/options.rb
+++ b/lib/kramdown/options.rb
@@ -245,6 +245,23 @@ Default: false
 Used by: HTML/Latex converter
 EOF
 
+    define(:disable_parsers, Object, [], <<EOF) do |val|
+List of parsers to disable
+
+Kramdown has a number of built-in span and block parsers, and
+additional parsers can be added by subclassing. This option allows the
+caller to disable specific parsers by name, especially when something
+like smart quotes or typographic dashes will be supplied by a separate
+processing step.
+
+Default: []
+Used by: kramdown parser
+EOF
+      val = simple_array_validator(val, :disable_parsers)
+      val.map! {|v| v.to_sym rescue v}
+      val
+    end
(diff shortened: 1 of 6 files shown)


gettalong:
This won't be merged since the kramdown parser is not a only-take-what-you-like parser, it is designed to parse all documents created for it and therefore isn't customizable.

If you don't like how the kramdown parser works and/or if you need to add or remove a certain parser, you can create a new parser based on the kramdown parser. See the Markdown parser for an example.

agriffis:
@gettalong Thanks for the hint! I ended up with a simple subclass, see https://github.com/scampersand/jekyll-pants/blob/master/lib/jekyll/pants/down.rb and instructions for using it with jekyll at https://github.com/scampersand/jekyll-pants#compatibility-with-kramdown

fliiiix:
A way to disable Typographic Symbols would be nice since there are user like me, who prefer ascii over typographic symbols. And it's a bit annoying to replace them back to ascii.

agriffis:
@fliiiix Understanding that you're probably looking for a command-line option or something... Nonetheless, the small subclass I provided at the link above does exactly what you're looking for.

PR #399, opened 2016-12-15
Fix Footnote link spacing

martynchamberlin:
There is a white space in the DOM from the end of the footnote to the beginning of the `<a>` tag that serves as a reversefootnote. I'd love to be able to replace this white space to `&nbsp;`. What this would effectively do is ensure that the backlink could never get pushed down to its own line in the footnote. If there wasn't enough room for it on the previous line, then since it'd be connected to the final word in the footnote, then it would push both the final word and the link to the next line. 

Perhaps illustrations would help. Here's the old way: 

And here's the new way: 

See how much better the latter looks?

### DIFF
--- a/lib/kramdown/converter/html.rb
+++ b/lib/kramdown/converter/html.rb
@@ -453,9 +453,9 @@ module Kramdown
           end
 
           unless @options[:footnote_backlink].empty?
-            para.children << Element.new(:raw, FOOTNOTE_BACKLINK_FMT % [insert_space ? ' ' : '', name, backlink_text])
+            para.children << Element.new(:raw, FOOTNOTE_BACKLINK_FMT % [insert_space ? '&nbsp;' : '', name, backlink_text])
             (1..repeat).each do |index|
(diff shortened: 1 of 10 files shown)


martynchamberlin:
I'll update the tests if we get the green light on this!

gettalong:
What if there are more than one backlinks? Should there also be a non-breaking space instead of a normal space?

martynchamberlin:
Ah good call. Yes, that should be a non-breaking space as well. I have changed both scenarios and also updated all of our test templates.

gettalong:
Thanks for the changes - I will include them with the next release! One more thing: Please merge your commits into one.

martynchamberlin:
@gettalong Excited to hear! I've squashed my second commit. Thanks.

martynchamberlin:
@gettalong Just curious when may we expect the next release? Sometime this Jan?

gettalong:
@martynchamberlin Yes, in the next few weeks when I have time.

gettalong:
Thanks for your contribution - will be released today.

martynchamberlin:
Lovely!

Issue #350, opened 2016-06-23
Rouge 2.0 compatibility


5t111111:
The recently released Rouge 2.0 has [breaking changes with its formatters](https://github.com/jneen/rouge#formatters).

I've checked the compatibility by providing those new formatters to Kramdown options and found some incompatibilities/strange behavior.

The issues I've found so far are:
- `Rouge::Formatters::HTMLLegacy` which should provide backwards-compatibility add `<pre>` tag to only the first code block.
- It seems you cannot specify formatters other than the default `Rouge::Formatters::HTML` because they require to be instantiated with base formatters.

Although I don't know if these issues are related Rouge internals rather than Kramdown at the moment, I just want to paste the test code before I can give it further investigation.

``````
require 'kramdown'
require 'rouge'

source = <<EOS
# Rouge 2.0 compatibility check

code block 1:

```ruby
def foo
  puts 'foo'
end
```

code block 2:

```ruby
def bar
  puts 'bar'
end
```

one liner:

```
def baz; puts 'baz'; end
```
EOS

# Without specifying a formatter
# It won't wrap code block with pre tag

puts Kramdown::Document.new(source, input: 'GFM', syntax_highlighter: 'rouge').to_html

# Specify Rouge::Formatters::HTML as a formatter
# It won't wrap code block with pre tag

puts Kramdown::Document.new(
  source,
  input: 'GFM',
  syntax_highlighter: 'rouge',
  syntax_highlighter_opts: {
    formatter: Rouge::Formatters::HTML
  }
).to_html

# Specify Rouge::Formatters::HTMLLegacy as a formatter
# It wrap only the first code block with pre tag

puts Kramdown::Document.new(
  source,
  input: 'GFM',
  syntax_highlighter: 'rouge',
  syntax_highlighter_opts: {
    formatter: Rouge::Formatters::HTMLLegacy
  }
).to_html

# Specify Rouge::Formatters::HTMLPygments as a formatter
# It causes an error

puts Kramdown::Document.new(
  source,
  input: 'GFM',
  syntax_highlighter: 'rouge',
  syntax_highlighter_opts: {
    formatter: Rouge::Formatters::HTMLPygments # => Causes an error because of missing base formatter
  }
).to_html
``````

gettalong:
kramdown supports rouge ~> 1.8 (see the Rakefile), so there is no real problem for kramdown because of rouge 2.x.

[...]

PR #413, opened 2017-02-22
Support Rouge 1.x and 2.0

5t111111:
This PR is kind of a draft proposal for supporting Rouge 2.0.

Since for the moment Kramdown supports Rouge < 2.0 which is considerably outdated, little by little it begins to hard to coexist with other gems which support Rouge >= 2.0 only.

I am still not sure what is the best way to support Rouge 2.0. Maybe we can support Rouge >= 2.0 only like other gems do (e.g. [middleman-syntax drops Rouge 1.x support](https://github.com/middleman/middleman-syntax/blob/270a86dd56220e069d55461ec570c73434de4a41/middleman-syntax.gemspec#L19)). However, I doubt that it is comfortable way for Kramdown because Rouge is an optional dependency and seem not to be easy to fix version to 2.0 or later.

So, this is the PR for supporting both Rouge 1.x and 2.0. Syntax highlighting will work just the same as before with Rouge 2.0, but it spoils some goodness, testability, for example.

How does this sound to you?

Related to https://github.com/gettalong/kramdown/issues/350

### DIFF
--- a/lib/kramdown/converter/syntax_highlighter/rouge.rb
+++ b/lib/kramdown/converter/syntax_highlighter/rouge.rb
@@ -17,6 +17,14 @@ module Kramdown::Converter::SyntaxHighlighter
 
       # Highlighting via Rouge is available if this constant is +true+.
       AVAILABLE = true
+
+      begin
+        # Rouge::Formatters::HTMLLegacy is available on Rouge 2.0 or later
+        FORMATTER_CLASS = ::Rouge::Formatters::HTMLLegacy
+      rescue NameError
+        # Fallbacks to Rouge 1.x formatter if Rouge::Formatters::HTMLLegacy is not available
+        FORMATTER_CLASS = ::Rouge::Formatters::HTML
+      end
     rescue LoadError, SyntaxError
       AVAILABLE = false  # :nodoc:
     end
@@ -26,8 +34,8 @@ module Kramdown::Converter::SyntaxHighlighter
       call_opts[:default_lang] = opts[:default_lang]
       lexer = ::Rouge::Lexer.find_fancy(lang || opts[:default_lang], text)
       return nil if opts[:disable] || !lexer
-
-      formatter = (opts.fetch(:formatter, ::Rouge::Formatters::HTML)).new(opts)
+      opts[:css_class] ||= 'highlight' # For backward compatibility when using Rouge 2.0
+      formatter = (opts.fetch(:formatter, FORMATTER_CLASS)).new(opts)
       formatter.format(lexer.lex(text))
     end
 
--- a/doc/syntax_highlighter/rouge.page
+++ b/doc/syntax_highlighter/rouge.page
@@ -11,6 +11,10 @@ To use Rouge, set the option ['syntax_highlighter'](../options.html#option-synta
 'rouge' and make sure that Rouge is available. The Rouge library can be installed, e.g., via
 Rubygems by running `gem install rouge`.
 
+Note that since Rouge 2.0 has breaking changes, Kramdown swaps formatter classes depending on
+Rouge version. By default, `Rouge::Formatters::HTMLLegacy` is used with Rouge 2.0 or later
+instead of `Rouge::Formatters::HTML` which is the default formatter with Rouge 1.x.
+
 The Rouge syntax highlighter supports the following keys of the option
 ['syntax_highlighter_opts'](../options.html#option-syntax-highlighter-opts):
 
@@ -18,7 +22,8 @@ default_lang
 : The default language that should be used when no language is set.
 
 formatter
-: A custom Rouge formatter class that should be used instead of `Rouge::Formatters::HTML`.
+: A custom Rouge formatter class that should be used instead of the following default formatter
+  (Rogue 1.x: `Rouge::Formatters::HTML` / Rogue 2.x: `Rouge::Formatters::HTMLLegacy`).
 
   Note that setting this key only makes sense using Ruby code because the value needs to be a class
   object!


lib/kramdown/converter/syntax_highlighter/rouge.rb
```diff
@@ -27,7 +27,16 @@ def self.call(converter, text, lang, type, call_opts)
...
 
-      formatter = (opts.fetch(:formatter, ::Rouge::Formatters::HTML)).new(opts)
+      begin
+        # Rouge::Formatters::HTMLLegacy is available on Rouge 2.0 or later
+        formatter_class = ::Rouge::Formatters::HTMLLegacy
+        opts[:css_class] ||= 'highlight'
+      rescue NameError
+        # Fallback to Rouge 1
+        formatter_class = ::Rouge::Formatters::HTML
+      end
```
gettalong: This `begin ... rescue ... end` block should go outside the method and inside the `begin ... rescue` block where we test for availability.

lib/kramdown/converter/syntax_highlighter/rouge.rb
```diff
@@ -27,7 +27,16 @@ def self.call(converter, text, lang, type, call_opts)
       lexer = ::Rouge::Lexer.find_fancy(lang || opts[:default_lang], text)
       return nil if opts[:disable] || !lexer
 
-      formatter = (opts.fetch(:formatter, ::Rouge::Formatters::HTML)).new(opts)
+      begin
+        # Rouge::Formatters::HTMLLegacy is available on Rouge 2.0 or later
+        formatter_class = ::Rouge::Formatters::HTMLLegacy
+        opts[:css_class] ||= 'highlight'
```
gettalong: Why are you assigning the 'highlight' CSS class here?

5t111111: This is requirement for backward compatibility.

While [Rouge < 2.0's used to add 'highlight' class to parsed codes](https://github.com/jneen/rouge/blob/0415a095744b46bad6c3aed35133458cc1770e2b/lib/rouge/formatters/html.rb#L27), Rouge >= 2.0 adds 'codehilite' by default even if you use `Rouge::Formatters::HTMLLegacy`. So, without specifing an explicit css class, users perhaps encounter regression.

lib/kramdown/converter/syntax_highlighter/rouge.rb
5t111111: I am not sure if I properly understand what you mean.

Do you mean that `begin ... rescue ... end` should be placed like [these ones](https://github.com/5t111111/kramdown/blob/d0e81c3ee1531e5723af8e9805cd33181f984665/lib/kramdown/converter/syntax_highlighter/rouge.rb#L15-L22) like the below snippet? 

```ruby

...

  module Rouge

...

    begin
      FORMATTER_CLASS = ::Rouge::Formatters::HTMLLegacy
    rescue NameError
      FORMATTER_CLASS = ::Rouge::Formatters::HTML
    end

    def self.call(converter, text, lang, type, call_opts)

      ...

      formatter = (opts.fetch(:formatter, FORMATTER_CLASS)).new(opts)
      formatter.format(lexer.lex(text))
    end

...

  end

```

lib/kramdown/converter/syntax_highlighter/rouge.rb
gettalong: Thanks for the explanation!

lib/kramdown/converter/syntax_highlighter/rouge.rb
gettalong: Yes, but directly into the other `begin ... rescue` block, right after `require 'rouge'`. This avoids doing this over and over again for each code block.

5t111111:
@gettalong Thanks for letting me know about your suggestion. Fixed!

lib/kramdown/converter/syntax_highlighter/rouge.rb
```diff
@@ -17,17 +17,23 @@ module Rouge
...
     end
 
     def self.call(converter, text, lang, type, call_opts)
       opts = options(converter, type)
       call_opts[:default_lang] = opts[:default_lang]
       lexer = ::Rouge::Lexer.find_fancy(lang || opts[:default_lang], text)
       return nil if opts[:disable] || !lexer
-
-      formatter = (opts.fetch(:formatter, ::Rouge::Formatters::HTML)).new(opts)
+      opts[:css_class] ||= 'highlight' # For backward compatibility when using Rouge 2.0
```
5t111111: Althogh this line actually does almost nothing when using Rouge < 2.0, I think it is better than checking a formatter's class in the method.

lib/kramdown/converter/syntax_highlighter/rouge.rb
```diff
@@ -17,17 +17,23 @@ module Rouge
 
       # Highlighting via Rouge is available if this constant is +true+.
       AVAILABLE = true
+
+      # Rouge::Formatters::HTMLLegacy is available on Rouge 2.0 or later
+      FORMATTER_CLASS = ::Rouge::Formatters::HTMLLegacy
     rescue LoadError, SyntaxError
       AVAILABLE = false  # :nodoc:
+    rescue NameError
```
gettalong: What I meant was putting the whole `begin ... rescue ... end` block inside the `begin...rescue` block where availability is checked. Then it is clear which exception handling mechanism handles which situation.

lib/kramdown/converter/syntax_highlighter/rouge.rb
gettalong: Agreed!

gettalong:
@5t111111 Some more things:

* Please update the rouge documentation at `doc/syntax_highlighter/rouge.page`
* Please update .travis.yml to make Travis test against rouge 1.x and 2.x (only if you know how to do this; otherwise I will do this later)
* Please squash your commits into one

Thanks!

5t111111:
- Moved new rouge 2.0 formatter availability check block into the block where rogue gem availability is checked
- Update  the rouge documentation at `doc/syntax_highlighter/rouge.page`
- Squash commits into one

> Please update .travis.yml to make Travis test against rouge 1.x and 2.x (only if you know how to do this; otherwise I will do this later)

Can you please do it later? I am not sure what is the appropriate way to archive this purpose.

Please point out to me if you have any further suggestinons.

gettalong:
Yes, I will do the Travis change later. Everything else looks great - thank you!

I will probably have some time this weekend to merge this in.

DaMouse404:
@gettalong any idea when this will get merged? 😄

gettalong:
@DaMouse404 It will definitely be in the next release but my time for open-source work is currently very limited. So, maybe sometime this month.

gettalong:
@5t111111 Thanks for you pull request - I have merged it and fixed the test runner and Travis.

PR #408, opened 2017-02-10
Small typo/grammar correction

suttonwilliamd:
Set the verb to past tense to match the sentence.

PR #421, opened 2017-03-20
Tiny grammar correction

suttonwilliamd:
This is a copy of #408, hopefully with less fail this time.


gettalong:
Thanks - I have added your commit!

Issue #172, opened 2014-10-06
add checkbox list to GFM 


fdschneider:
To create ToDo lists I love GitHubs checkbox list syntax:

```
- [x] this is still to do
- [ ] done 
```

renders as
- [ ] this is still to do
- [x] done 

I would like to use this on a Jekyll page. Also, kramdowns GFM parser (#68) is [fairly incomplete](https://help.github.com/articles/github-flavored-markdown/), so it would be great to include it. Anybody out there who is interested in implementing this?

zerocool4u2:
you can just replace the text before passing it to the kramdown parser
```ruby
raw_text.gsub(/(\d+\. |\* )\K\[(x|X)\]/, %(<i class="fa fa-check-square-o"></i>)).gsub(/(\d+\. |\* )\K\[ \]/, %(<i class="fa fa-square-o"></i>))
```
*FontAwesome example*

`\K` will work only on ruby 2.0 and above, it's like a lookbehind, but will work with the + and | (or) escape chars, keep in mind that will also replace the code on code blocks, i guess you could use a more complex regexp if that is an issue
http://rubular.com/r/wvxaX8jKtX
http://rubular.com/r/5hMGNPFl87

alequartz:
@gettalong  I am following the [gitlab discussions](https://gitlab.com/gitlab-org/gitlab-ce/issues/18552) on this topic. Apparently they would be willing to switch to kramdown if the checkbox lists are implemented. I cannot hide that I am totally in favour.
Would the solution from @zerocool4u2 work for you? Are you still open to pull requests on this ? 

Thank you for all your work

gettalong:
@alequartz Yes, pull requests are welcome. Note, however, that this can only be an enhanced to the GFM parser, not to the main kramdown parser.

The first thing to be done would be to define the required output. Github *seems* to output this for the example of @fdschneider:

~~~ html
<ul class="contains-task-list">
<li class="task-list-item"><input type="checkbox" class="task-list-item-checkbox" disabled=""> this is still to do</li>
<li class="task-list-item"><input type="checkbox" class="task-list-item-checkbox" checked="" disabled=""> done</li>
</ul>
~~~

So is this compatible with the way Gitlab renders task lists?

alequartz:
@gettalong. This is what I get using the same syntax in gitlab. Pretty similar. Except for the ul class. I suggest we stick with the github class.

```html
<ul class="task-list" dir="auto">
   <li class="task-list-item">
[...]

PR #442, opened 2017-07-14
Added task list ability to GFM parser.

adwylie:
Followed most of the suggestions in #172. Never written Ruby before though so it may be a bit sloppy.

Currently renders as li disc followed by the checkbox, was assuming there'd be css to change `list-style-type` to `none` if the checkbox exists.  If not I can change it.

Thanks, Andrew.

### DIFF
--- a/lib/kramdown/parser/gfm.rb
+++ b/lib/kramdown/parser/gfm.rb
@@ -158,25 +158,24 @@ module Kramdown
       # elements where necessary (as well as applying classes to the ul/ol and li elements).
       def parse_list
         super
-        @tree.children.each do |element|
-          if [:ul, :ol].include? element.type
-            is_tasklist = false
-            element.children.each do |li|
-              # li -> p -> raw_text
-              checked = li.children[0].children[0].value.gsub!(
-                  /\[ \]\s+/, '<input type="checkbox" class="task-list-item-checkbox" disabled="disabled" />')
-              unchecked = li.children[0].children[0].value.gsub!(
-                  /\[x\]\s+/i, '<input type="checkbox" class="task-list-item-checkbox" disabled="disabled" checked="checked" />')
-              is_tasklist = (checked != nil or unchecked != nil)
-              if is_tasklist
-                li.attr[:class] = 'task-list-item'
-              end
-            end
-            if is_tasklist
-              element.attr[:class] = 'task-list'
-            end
-          end
+        current_list = @tree.children.select{ |element| [:ul, :ol].include?(element.type) }.last
+
+        is_tasklist = false
+        box_unchecked = '<input type="checkbox" class="task-list-item-checkbox" disabled="disabled" />'
+        box_checked = '<input type="checkbox" class="task-list-item-checkbox" disabled="disabled" checked="checked" />'
+
+        current_list.children.each do |li|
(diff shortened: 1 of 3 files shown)


lib/kramdown/parser/gfm.rb
```diff
@@ -154,6 +154,31 @@ def parse_strikethrough_gfm
         el
       end
 
+      # To handle task-lists we override the parse method for lists, converting matching text into checkbox input
+      # elements where necessary (as well as applying classes to the ul/ol and li elements).
+      def parse_list
+        super
+        @tree.children.each do |element|
```
gettalong: Using `@tree.children.each` would mean that, e.g., parsing the third list involves checking the first two lists. It would be better to restrict the processing to the list at hand which saves time.

lib/kramdown/parser/gfm.rb
```diff
@@ -154,6 +154,31 @@ def parse_strikethrough_gfm
...
+          if [:ul, :ol].include? element.type
```
gettalong: Please use parentheses around method arguments.

lib/kramdown/parser/gfm.rb
```diff
@@ -154,6 +154,31 @@ def parse_strikethrough_gfm
...
+            is_tasklist = false
+            element.children.each do |li|
+              # li -> p -> raw_text
+              checked = li.children[0].children[0].value.gsub!(
+                  /\[ \]\s+/, '<input type="checkbox" class="task-list-item-checkbox" disabled="disabled" />')
+              unchecked = li.children[0].children[0].value.gsub!(
+                  /\[x\]\s+/i, '<input type="checkbox" class="task-list-item-checkbox" disabled="disabled" checked="checked" />')
+              is_tasklist = (checked != nil or unchecked != nil)
```
gettalong: Please use `||` instead of `or`; and `checked.nil?` as this avoids a method call.

Since `is_tasklist` is also used outside this loop, you need to use `is_tasklist ||= ...` since otherwise only the state of the last list item determines the status (which is probably not wanted).

lib/kramdown/parser/gfm.rb
```diff
@@ -154,6 +154,31 @@ def parse_strikethrough_gfm
...
+              if is_tasklist
+                li.attr[:class] = 'task-list-item'
```
gettalong: * Please use the if-modifier syntax syntax for this since it is a short line and easy to understand.
* Use `attr['class']` since the keys are strings, not symbols.

All this also applies to line 176.

adwylie:
Hey. I've addressed all of the comments and added some tests. Thanks!

gettalong:
Thanks - I'm currently rather busy but will have a look at soon as possible!

gettalong:
Thanks for your contribution  - I have squashed the commits and integrated your code.

robbiejaeger:
@adwylie @gettalong Thanks for this added feature!

Curious what the decision was for making the checkbox disabled by default?

gettalong:
@robbiejaeger kramdown generates static content, so if the input would be clickable the visible state would change but nothing else.

robbiejaeger:
Makes sense - thanks for the clarification!

PR #448, opened 2017-09-11
Allow for passing Rouge formatter as a `String`

kejadlen:
I thought this might be handy so that Jekyll users can change the Rouge formatter.


gettalong:
Thanks for your pull request but I'm not really convinced that this is a good change... will have to think about it.

kejadlen:
Yeah, I can see that - do you suppose the `String` to `Class` transform might make more sense in Jekyll than Kramdown? I'm not too familiar with either project, so don't know what's reasonable.

gettalong:
@kejadlen Thanks for your pull request! After some consideration, I have changed your pull request a bit (remove unnecessary changes, changed `String` to a more restricted regex) and included it.

I.e. this will be in the next release.

kejadlen:
Cool, thanks!

PR #457, opened 2017-09-23
Adding RPM installation instructions

bexelbie:


gettalong:
I will consider adding a method description to installing Rubygems on non-deb systems but not kramdown itself since the packaged versions are often out of date.

Also, if another method is described, the text above needs to be changed accordingly.

ccorn:
On RPM-based CentOS-7, the package manager command would rather be `yum` instead of `dnf`. And yes, distro-based ruby packages are typically outdated (kramdown version is 1.9 for CentOS-7). Moreover those are often split e.g. into `rubygem-kramdown` and `rubygem-kramdown-doc`.

I suppose that the demonstrated package manager command is generally of little significance. Those who have the admin rights to install packages already know the basic package manager commands for their distribution, or they use a GUI frontent and need to know the package name only. In the existing guide, the relevant package name is `ruby` because the remainder is done via `gem` directly.

However, in no distro I would `gem install` in system directories maintained by the distro's package manager. That would invite conflicts. I have a MacPorts version of Ruby installed  (`sudo port install ruby24; sudo port select --set ruby ruby24`), but the gems of my interest I install in my home directory to have easier access to updates. Therefore I would rather suggest
~~~ bash
cat >~/.gemrc <<EOF
install: --user-install
update: --user-install
uninstall: --user-install
EOF
~~~
and thereafter a simple
~~~ bash
gem install kramdown
~~~
does the job, notably without `sudo`.
Giving the explicit command to install Ruby is probably too much potentially incorrect detail, but the per-user gem handling is both useful and easier to get universally right.

bexelbie:
@gettalong I respect not wanting to rely on distributions having out of date code, however most distributions do a very good job and b) then maybe you shouldn't mention debian either?

I think the biggest thing we want to bring to people is not the actual install command, but the package name.

gettalong:
The problem with not using `sudo` when running `gem install kramdown` is that the binary is not available in the PATH.

As for whether to include Ubuntu but not Fedora/...: As stated on the page it is just an example and I chose Ubuntu because a) I use it and b) it is the most common one.

So which package would be required on an RPM system (OpenSUSE, RedHat, Fedora, CentOS) to run `sudo gem install kramdown`?

ccorn:
On a CentOS 7 system that I have access to, the `gem` command comes packaged as `rubygems` and is automatically installed as a dependency when you `sudo yum install ruby`.
Since `gem install` may need to compile something for a dependency gem, `gem` users should install `ruby-devel` as well, probably along with `gcc` and `glibc-devel`.
As noted above, a kramdown package is available as `rubygem-kramdown` and `rubygem-kramdown-doc` (via the `epel` repository), but at version 1.9.

gettalong:
@ccorn Thanks for your input! However, I have reworded the installation instruction and will not put any mention of yum/dnf/zypper/... in there because I think that would just be overkill.

Issue #459, opened 2017-10-11
[GFM] wrong ID generation for headers containing --


kolyshkin:
Consider the following fragment of an `.md` file on github:

```md
...
see [--foobar](#option---foobar) option below.
...
## Option --foobar
```

The above gets rendered by GitHub to the following:

...
see [--foobar](#option---foobar) option below.
...
## Option --foobar

Now, if we use kramdown GFM to convert this .md, we'll have the following two problems:

1. Double dash gets replaced by ndash, which is bad at least in this particular case, and maybe in general, as GitHub's own parser doesn't do it.
2. The link to "Option --foobar" will not work, as ID for the intralink will be `option-foobar` rather than github's `option---foobar`. The reason is the above double-dash -> ndash conversion.

While the first issue is only bad in some specific cases, the second one is really bad as it breaks some intralinks in the resulting document. Perhaps it makes sense to disable double-dash and triple-dash conversion for GFM (say, another option to gfm_quirks)?

gettalong:
If we change the way link generation works for the GFM parser, it might break existing documents. However, since it is GFM we are talking about and the implementation in kramdown is not quite complete, I would consider such a change a bug fix and not a feature change.

A pull request for the this is welcome.

kolyshkin:
OK, I have a PR with a patch that apparently works. I'm just not sure about it as my Ruby knowledge is totally non-existent, to say at least.

gettalong:
Pull request #462 is applied.

PR #462, opened 2017-10-17
GFM parser: option to disable typographic symbol conversion

kolyshkin:
With GFM, we want to be as close as possible to GitHub Flavored Markdown. In particular, GitHub does not replace `--` with en-dash, `---` with em-dash et cetera, and doing this not only breaks HTML rendering (resulting in, say, `–option` instead of `--option`), but also makes header IDs incompatible with those generated by GitHub. So, let's turn automatic typographic symbol conversion by default, and add an option to enable it back.

I have checked that this change makes header IDs the same as they are in GitHub.

For more details on the issue and examples, see https://github.com/gettalong/kramdown/issues/459

Signed-off-by: Kir Kolyshkin <kolyshkin@gmail.com>

### DIFF
--- a/lib/kramdown/options.rb
+++ b/lib/kramdown/options.rb
@@ -614,6 +614,13 @@ separated by commas. Possible names are:
   Note that if this quirk is used, lazy line wrapping does not fully
   work anymore!
 
+* no_auto_typographic
+
+  Disables automatic conversion of some characters into their
+  corresponding typographic symbols (like `--` to em-dash etc).
+  This helps to achieve results closer to what GitHub Flavored
+  Markdown produces.
+
 Default: paragraph_end
 Used by: GFM parser
 EOF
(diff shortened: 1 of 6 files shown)


gettalong:
Thanks for your pull request!

Some comments:

* Please use the "gfm_quirks" configuration option instead of introducing a new one.
* The default value should not change, i.e. this quirk should not automatically applied.

kolyshkin:
> * Please use the "gfm_quirks" configuration option instead of introducing a new one.
> * The default value should not change, i.e. this quirk should not automatically applied.

Done. I have two concerns though:

1. It looks like the value of `gfm_quirks` is not validated. Is it intentional (i.e. non-recognized options are silently ignored)? To my mind, it would make sense to have at least a warning printed, to figure out a typo or, say, an old version of parser where a specific option is not recognized.

2. If this option/quirk is not enabled by default, we still have those two problems that I described in the https://github.com/gettalong/kramdown/issues/459. Isn't it your intention to make GFM parser as compatible with GFM as possible?

gettalong:
Thanks for the changes! As for your concerns:

1. Yes, this is intentional so that custom parsers/converter can reuse the option.

2. Full compatibility with GFM will not be possible, at first because there was no spec and now because it is based on CommonMark. Therefore having this as option is preferable.

gettalong:
I have cherry-picked your commit - the changes will be in the next release.

PR #470, opened 2017-11-13
Correctly convert `img` tags without a `src` attribute

donv:
`img` tags without a `src` attribute would give an "undefined method `empty?' for nil:NilClass" exception.


gettalong:
Thank you for your contribution! I have squashed your commits and included them for the next release.

PR #479, opened 2017-12-08
Removed pdf-as-output from man page

gonsie:
Kramdown support PDF output via LaTeX, it cannot natively generate a PDF.
Currently, passing `-o pdf` to the command line tool generates an error.


gettalong:
kramdown *does* support direct PDF output but the converter may not support all features.

For example, the following just works:

~~~
echo -e "# header\n\nsome para\n\n* a list\n* item\n" | rl bin/kramdown -o pdf > /tmp/o.pdf
~~~

PR #484, opened 2018-01-04
Expanded ToC example

milosimpson:
Expanded the ToC doc to make it more clear that the {:toc} has to be used in conjunction with a markdown list, i.e. {:toc} is not a "valid" expression all by itself.

Ala https://stackoverflow.com/questions/48071478/table-of-contents-in-markdown-using-only-the-standard-out-of-the-box-github-p


gettalong:
Thanks for your contribution - I have applied it.

PR #487, opened 2018-01-22
reverse cap and cup

kivikakk:
See https://dev.w3.org/html5/html-author/charref; `&cap;` is `&#8745`; and `&cup;` is `&#8746;`.


gettalong:
Thanks!

PR #493, opened 2018-02-04
KaTeX and SsKaTeX: Usage notes and comparison

ccorn:
As promised, some usage notes and guidelines on where and whether to use KaTeX or SsKaTeX.

### DIFF
--- a/doc/math_engine/katex.page
+++ b/doc/math_engine/katex.page
@@ -5,10 +5,24 @@ title: KaTeX
 ## Math Engine KaTeX
 
 This math engine uses [KaTeX] via the [katex rubygem][katex-gem] to convert TeX math formulas into
-KaTeX HTML (using [ExecJS] under the hood).
+KaTeX HTML, using [ExecJS] under the hood. This eliminates the need for client-side math-rendering
+Javascript. This engine is somewhat similar to the [SsKaTeX] math engine, but easier to use and
+suitable for kramdown input from untrusted users.
 
-This eliminates the need for client-side math-rendering JavaScript. Your HTML templates still need
-to include the KaTeX CSS (see the [gem's documentation][katex-gem#assets] on this).
+Requirements for running kramdown with math engine KaTeX:
+
+- Ruby gem [`katex`][katex-gem],
+- Ruby gem [`execjs`][ExecJS],
+- A Javascript engine supported by ExecJS, e.g. via one of
+  - Ruby gem [`therubyracer`][therubyracer],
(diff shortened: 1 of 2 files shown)


gettalong:
Thanks - I have merged your changes!

doc/math_engine/katex.page
```diff
@@ -5,10 +5,24 @@ title: KaTeX
...
+KaTeX HTML, using [ExecJS] under the hood. This eliminates the need for client-side math-rendering
+Javascript. This engine is somewhat similar to the [SsKaTeX] math engine, but easier to use and
+suitable for kramdown input from untrusted users.
 
-This eliminates the need for client-side math-rendering JavaScript. Your HTML templates still need
-to include the KaTeX CSS (see the [gem's documentation][katex-gem#assets] on this).
+Requirements for running kramdown with math engine KaTeX:
+
+- Ruby gem [`katex`][katex-gem],
+- Ruby gem [`execjs`][ExecJS],
```
glebm: This is a `runtime_dependency` of the gem so there is no need to install it separately:
https://github.com/glebm/katex-ruby/blob/e432853a5ae1122e8dc8ce96147bded41494baef/katex.gemspec#L25

doc/math_engine/katex.page
```diff
@@ -5,10 +5,24 @@ title: KaTeX
...
+- A Javascript engine supported by ExecJS, e.g. via one of
+  - Ruby gem [`therubyracer`][therubyracer],
```
glebm: `therubyracer` is not a good choice for a runtime, `mini_racer` is better in every way.

ccorn: Those are just examples of what is supported by ExecJS. I have chosen `therubyracer` mostly because of the naming similarity to `therubyrhino` which is suitable for Jruby.

doc/math_engine/katex.page
ccorn: That's just saying that the requirement usually gets fulfilled by dependency resolvers, but it's still a requirement. And it informs about what will (have to) be found in a functioning installation.

doc/math_engine/katex.page
ccorn: Besides, ExecJS has a different view. Its autodetection prefers `RubyRacer` over `MiniRacer`, trying `RubyRhino` and `Duktape` in between. I do not suppose that that is an argument by itself; my actual point is that one should not interpret too much into that list of examples, because there is no way to satisfy all tastes anyway.

PR #499, opened 2018-03-04
Rcfile options and their application to test output updates

ccorn:
This short sequence of commits introduces
1. New CLI-only options `-n`, `--no-rcfile` and `-r` *FILE*, `--rcfile` *FILE* that control whether to load a `kramdownrc` file as well as its pathname. The `-n` ensures that no `kramdownrc` can influence the output of your invocation of `bin/kramdown`, and the `-r` *FILE* replaces the default configuration with that contained in *FILE*. With those, generating test reference outputs from the command line can be done as follows:

   ~~~ shell
   ruby -Ilib bin/kramdown -r mytest.options mytest.text >mytest.html
   ~~~
   Note that `mytest.options` does not need to exist. As `kramdownrc` is optional, so is the file named with `-r`.
2. An update to `Rakefile` that makes use of those options. The existing targets `dev:test_*_deps` and `dev:update_*_tests` have made use of `bin/kramdown` before; those invocations are now protected from `kramdownrc` influence and look more like the example above. Also, instead of manually enumerating test file names in the `Rakefile`, globs are used now.
3. New `Rakefile` targets for the MathjaxNode engine similar to those for SsKaTeX and KaTeX. Concretely, a new target `dev:update_mathjaxnode_tests` and its dependency `dev:test_mathjaxnode_deps` which does some sanity checking as to whether MathjaxNode actually works.

With those, the recent MathjaxNode test reference output updates could have been accomplished with the command

~~~ shell
rake dev:update_mathjaxnode_tests
~~~
And the output (if a working mathjax-node-cli is installed) would be

~~~ shell-session
NO RELEASE NOTES/CHANGES FILE
MathjaxNode is available, and its default configuration works.
/home/ccorn/.rvm/rubies/ruby-2.4.2/bin/ruby -Ilib bin/kramdown -r test/testcases/span/math/mathjaxnode.options test/testcases/span/math/mathjaxnode.text >test/testcases/span/math/mathjaxnode.html.19
/home/ccorn/.rvm/rubies/ruby-2.4.2/bin/ruby -Ilib bin/kramdown -r test/testcases/block/15_math/mathjaxnode_notexhints.options test/testcases/block/15_math/mathjaxnode_notexhints.text >test/testcases/block/15_math/mathjaxnode_notexhints.html.19
/home/ccorn/.rvm/rubies/ruby-2.4.2/bin/ruby -Ilib bin/kramdown -r test/testcases/block/15_math/mathjaxnode.options test/testcases/block/15_math/mathjaxnode.text >test/testcases/block/15_math/mathjaxnode.html.19
/home/ccorn/.rvm/rubies/ruby-2.4.2/bin/ruby -Ilib bin/kramdown -r test/testcases/block/15_math/mathjaxnode_semantics.options test/testcases/block/15_math/mathjaxnode_semantics.text >test/testcases/block/15_math/mathjaxnode_semantics.html.19
~~~

### DIFF
--- a/bin/kramdown
+++ b/bin/kramdown
@@ -27,11 +27,7 @@ begin
 rescue
 end
 
-options = if config_file && File.exist?(config_file)
-            YAML.safe_load(File.read(config_file), [Symbol])
-          else
-            {}
-          end
+options = {}
 format = ['html']
 
 OptionParser.new do |opts|
@@ -46,7 +42,14 @@ OptionParser.new do |opts|
           "html, GFM or markdown") {|v| options[:input] = v}
   opts.on("-o", "--output ARG", Array, "Specify one or more output formats separated by commas: " \
           "html (default),", "kramdown, latex, pdf, man or remove_html_tags") {|v| format = v}
-
+  opts.separator ""
+  opts.on("--no-config-file", "Do not read any configuration file. Default behavior is to check " \
+          "for a", "configuration file and read it if it exists.") {config_file = nil}
+  opts.on("--config-file FILE", "Specify the name of a configuration file with kramdown options " \
+          "in YAML", "format, e.g. \"auto_id_prefix: ARG\" instead of \"--auto-id-prefix ARG\"",
+          "and \"auto_ids: false\" instead of \"--no-auto-ids\".",
+          "Default: #{config_file}") {|v| config_file = v}
+  opts.separator ""
   opts.on("-v", "--version", "Show the version of kramdown") do
     puts Kramdown::VERSION
     exit
@@ -79,6 +82,17 @@ OptionParser.new do |opts|
 end.parse!
 
 begin
+  if config_file && File.exist?(config_file)
+    config_file_options = YAML.safe_load(File.read(config_file), [Symbol])
+    case config_file_options
+    when nil  # empty configuration file except perhaps YAML header and comments
+      # Nothing to do
+    when Hash
+      options = config_file_options.merge(options)
+    else
+      raise Kramdown::Error, "No YAML map in configuration file \"#{config_file}\""
+    end
+  end
   doc = Kramdown::Document.new(ARGF.read, options)
   result = ''
   format.each {|f| result = doc.send("to_#{f}")}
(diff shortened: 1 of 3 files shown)


bin/kramdown
```diff
@@ -46,6 +42,9 @@ OptionParser.new do |opts|
           "html, GFM or markdown") {|v| options[:input] = v}
   opts.on("-o", "--output ARG", Array, "Specify one or more output formats separated by commas: " \
           "html (default),", "kramdown, latex, pdf, man or remove_html_tags") {|v| format = v}
+  opts.on("-n", "--no-rcfile", "Do not read any configuration file") {config_file = nil}
+  opts.on("-r", "--rcfile FILE", "Read specified configuration file, if it exists.",
+          "Default: #{config_file}") {|v| config_file = v}
```
gettalong: I would go with `--config-file FILE` and `--no-config-file`, no short codes. And mention in `--no-config-file` that the default is using it if it is available. Should also go into the man page.

bin/kramdown
```diff
@@ -79,6 +78,17 @@ OptionParser.new do |opts|
 end.parse!
 
 begin
+  if config_file && File.exist?(config_file)
+    rcopts = YAML.safe_load(File.read(config_file), [Symbol])
```
gettalong: Please use `config_file_options` or something similar instead of `rcopts`.

ccorn:
Done as requested. Notes:

* I have taken the liberty to add a little more explanation of the option naming in the `--help` output and in the manpage. The explanations are worded differently because the manpage has some useful introduction that the `--help` output has not.
* Without short option names, the `--help` output looked as if `--no-config-file` somehow belonged to the description of `-o`. I have added separator gaps to avoid that impression. If you find that still out of balance, consider adding three more separator gaps to separate all option descriptions without exception, as in the manpage.
* Besides: The `-o` description is a bit too wide for an 80-column terminal; breaking the line after `commas:` would look better. Not included here because it's off-topic.
* Changed the comments for the `nil` YAML case to be more precise. I should mention that the treatment of `nil` is a change to previous behavior; the rationale being that empty configuration files should be considered valid.

ccorn:
ust noticed that the revised commit dates are wrong; seems to be an artifact of rebasing. I assure you that I have not foreseen your requests two days earlier :-)

gettalong:
Thank you for your changes!

PR #501, opened 2018-03-16
Minor optimizations with faster alternatives

ashmaroli:
- `Range#cover?` is [faster](https://github.com/JuanitoFatas/fast-ruby#cover-vs-include-code) than `Range#include?`
- `String#gsub` is [slower](https://github.com/JuanitoFatas/fast-ruby#stringgsub-vs-stringtr-code) than `String#tr` (for like numbered replacements)
- `String#gsub regex` is slower than `String#gsub str`
- Prefer native methods over regexp when possible


gettalong:
Thanks but did you measure the performance difference? As far as I can determine without benchmarking these changes will probably have no measurable impact.

ashmaroli:
> these changes will probably have no measurable impact.

Yes, you're correct. The *minor* gains are swallowed up somewhere else..

ashmaroli:
> reopen it if the performance gains are measurable

Ah.. Okay..

PR #502, opened 2018-03-16
Ruby versions older than 2.0 are not supported

ashmaroli:
Teardown checks for unsupported Ruby versions in `lib/` files


gettalong:
Thanks - I have cherry-picked your commit!

PR #510, opened 2018-05-17
Update katex dev dependency and tests

glebm:
/cc @ccorn


ccorn:
Looks good to me. Works with Duktape and Node (at least).
I have also checked that some rendering bugs from the prereleases KaTeX-0.9.0rc* are fixed (at least with Firefox). The stable release seems OK so far.
The only downside is that the new KaTeX version seems 60% slower. It's still fast, but 0.8.3 is significantly faster.

gettalong:
Thanks - I have cherry-picked your commit!

PR #512, opened 2018-05-17
Frozen string literals throughout

glebm:
This will be the default in Ruby 3.
This change is *dangerous*, as there may be cases not covered by tests.


gettalong:
Thanks for this pull request but I have to decline at the moment. I may change kramdown to use frozen strings in the future but not right now.

plentz:
@gettalong do you think that this is a good time to start changing things to use frozen string literals?

gettalong:
No, as @glebm wrote: It's a big change and right now I don't have the time.

plentz:
hum, ok. When you think it's a good time, ping me - I can offer some help.

glebm:
To save time, consider merging this and releasing a pre-release gem, e.g. `1.18.0.beta`. 

Then wait for a couple of months to see if any bugs come up.

gettalong:
@glebm FYI frozen string literals are now active with the latest commits.

PR #513, opened 2018-05-18
Optimize header parsing

### LINKED ISSUE #505: Performance issue when rendering a lot of space
Here's a pretty comprehensive example:

```ruby
note = "1"+" "*20000+"2"
Benchmark.measure { Kramdown::Document.new(note) }.real
```

This takes around 3 seconds on my machine. A DOS vulnerability potentially?

glebm:
Fixes #505

Benchmarks:

"setext":

```bash
ruby -rbenchmark -Ilib -rkramdown -e 'p Benchmark.measure{Kramdown::Document.new("1#{" "*20000}2\n==\n")}'
```

"atx":

```bash
ruby -rbenchmark -Ilib -rkramdown -e 'p Benchmark.measure{Kramdown::Document.new("## 1#{" "*20000}2")}'
```

### DIFF
--- a/lib/kramdown/parser/gfm.rb
+++ b/lib/kramdown/parser/gfm.rb
@@ -112,22 +112,16 @@ module Kramdown
         @options[:auto_id_prefix] + result
       end
 
-      ATX_HEADER_START = /^\#{1,6}\s/
+      ATX_HEADER_START = /^(?<level>\#{1,6})[\t ]+(?<contents>.*)\n/
       define_parser(:atx_header_gfm, ATX_HEADER_START, nil, 'parse_atx_header')
       define_parser(:atx_header_gfm_quirk, ATX_HEADER_START)
 
       # Copied from kramdown/parser/kramdown/header.rb, removed the first line
       def parse_atx_header_gfm_quirk
-        start_line_number = @src.current_line_number
-        @src.check(ATX_HEADER_MATCH)
-        level, text, id = @src[1], @src[2].to_s.strip, @src[3]
+        text, id = parse_header_contents
+        text.sub!(/[\t ]#+\z/, '') && text.rstrip!
         return false if text.empty?
-
-        @src.pos += @src.matched_size
-        el = new_block_el(:header, nil, nil, :level => level.length, :raw_text => text, :location => start_line_number)
-        add_text(text, el)
-        el.attr['id'] = id if id
-        @tree.children << el
+        add_header(@src["level"].length, text, id)
         true
       end
 
--- a/lib/kramdown/parser/kramdown/header.rb
+++ b/lib/kramdown/parser/kramdown/header.rb
@@ -43,7 +43,7 @@ module Kramdown
 
       HEADER_ID = /[\t ]{#(?<id>[A-Za-z][\w:-]*)}\z/
 
-      # @return [[String, String]] header text and optional ID.
+      # Returns header text and optional ID.
       def parse_header_contents
         text = @src["contents"]
         text.rstrip!
@@ -56,9 +56,6 @@ module Kramdown
         [text, id]
       end
 
-      # @param [Number] level
-      # @param [String] text
-      # @param [String, nil] id
       def add_header(level, text, id)
         start_line_number = @src.current_line_number
         @src.pos += @src.matched_size
--- a/lib/kramdown/parser/kramdown/header.rb
+++ b/lib/kramdown/parser/kramdown/header.rb
@@ -13,47 +13,60 @@ module Kramdown
   module Parser
     class Kramdown
 
-      HEADER_ID=/(?:[ \t]+\{#([A-Za-z][\w:-]*)\})?/
-      SETEXT_HEADER_START = /^(#{OPT_SPACE}[^ \t].*?)#{HEADER_ID}[ \t]*?\n(-|=)+\s*?\n/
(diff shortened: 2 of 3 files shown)


lib/kramdown/parser/gfm.rb
```diff
@@ -112,22 +112,16 @@ def generate_gfm_header_id(text)
...
       define_parser(:atx_header_gfm, ATX_HEADER_START, nil, 'parse_atx_header')
       define_parser(:atx_header_gfm_quirk, ATX_HEADER_START)
 
       # Copied from kramdown/parser/kramdown/header.rb, removed the first line
       def parse_atx_header_gfm_quirk
-        start_line_number = @src.current_line_number
-        @src.check(ATX_HEADER_MATCH)
-        level, text, id = @src[1], @src[2].to_s.strip, @src[3]
+        text, id = parse_header_contents
+        text.sub!(/[\t ]#+\z/, '') && text.rstrip!
```
gettalong: Is there a reason why these two statements are on one line? I would rather prefer them on separate lines.

lib/kramdown/parser/kramdown/header.rb
```diff
@@ -13,47 +13,60 @@ module Kramdown
   module Parser
     class Kramdown
 
-      HEADER_ID=/(?:[ \t]+\{#([A-Za-z][\w:-]*)\})?/
-      SETEXT_HEADER_START = /^(#{OPT_SPACE}[^ \t].*?)#{HEADER_ID}[ \t]*?\n(-|=)+\s*?\n/
+      SETEXT_HEADER_START = /^#{OPT_SPACE}(?<contents>.*)\n(?<level>[-=])[-=]*[ \t\r\f\v]*\n/
```
gettalong: Why did you leave out `[^ \t]` in the new version? It is not strictly necessary for the main kramdown parser due to the fixed order of block parsers but behaviour of derived parser may change. As far as I can tell, it doesn't make a performance difference.

Also: What is the reason behind changing `\s` to `[ \t\r\f\v]`?

lib/kramdown/parser/kramdown/header.rb
```diff
@@ -13,47 +13,60 @@ module Kramdown
...
+        add_header(@src["level"].length, text, id)
+        true
+      end
+      define_parser(:atx_header, ATX_HEADER_START)
+
+      protected
+
+      HEADER_ID = /[\t ]{#(?<id>[A-Za-z][\w:-]*)}\z/
+
+      # @return [[String, String]] header text and optional ID.
```
gettalong: Please use standard RDoc comments.

lib/kramdown/parser/kramdown/header.rb
```diff
@@ -13,47 +13,60 @@ module Kramdown
...
+          id = id_match["id"]
+          text = text[0...-id_match[0].length]
+          text.rstrip!
+        end
+        [text, id]
+      end
 
+      # @param [Number] level
+      # @param [String] text
+      # @param [String, nil] id
```
gettalong: Please use standard RDoc comments.

lib/kramdown/parser/kramdown/header.rb
glebm: I'm not familiar with RDoc, and although I've read the [RDoc Markup Reference](https://ruby.github.io/rdoc/RDoc/Markup.html), it doesn't seem to mention the correct markup for type hints, so I'm not sure what to do here.

The format I've used is YARD, which is also widely supported by IDEs for type hints. Can you convert to the desired format when merging?

lib/kramdown/parser/gfm.rb
glebm: There is no need to call `rstrip!` if `sub!` return nil.

The other options here are:

A:

```
text.rstrip! if text.sub!(...)
```

B:

```
if text.sub!(...)
  text.rstrip!
end
```

I think the current one (`&&`) is quite readable but can go with `A` or `B` if that's what you prefer.

lib/kramdown/parser/kramdown/header.rb
glebm: > Why did you leave out [^ \t] in the new version?

Because it's not necessary as we have to validate header length later on anyway. I have now added it back in. No performance difference.

> Also: What is the reason behind changing \s to [ \t\r\f\v]?

I find constructs like `\s*?\n` a bit cryptic. Performance-wise both are the same.

lib/kramdown/parser/gfm.rb
gettalong: Personally I would just put them on separate lines and take this minimal performance hit because it would be easier to see what's going on without the additional conditional.

lib/kramdown/parser/kramdown/header.rb
glebm: Done

lib/kramdown/parser/kramdown/header.rb
glebm: Done

gettalong:
I just noticed that the PR doesn't include the test cases themselves. Could you add them or should I add them after I merge the changes?

And please squash the commits into one before I merge them - thank you!

glebm:
> I just noticed that the PR doesn't include the test cases themselves. Could you add them or should I add them after I merge the changes?

Please add them yourself, it will be faster this way.

> And please squash the commits into one before I merge them - thank you!

Nowadays you can do it yourself at merge time by clicking "Squash and merge"!

gettalong:
Merged - thanks!

krasnoukhov:
@gettalong Will you cut a release please? Thank you

PR #518, opened 2018-05-24
Add more auto-generated files to .gitignore

glebm:


.gitignore
```diff
@@ -1,3 +1,10 @@
 htmldoc
 pkg
 webgen-tmp
+.bundle/
+CONTRIBUTERS
+VERSION
+kramdown.gemspec
+man/man1/kramdown.1
+node_modules/
+package-lock.json
```
gettalong: Lines 4, 9 and 10 have nothing to do with kramdown, please remove them. The other changes are fine.

glebm: Removed line 4.
Line 9 and 10 are for `npm install`:

https://github.com/gettalong/kramdown/blob/55d407290b62b1431e9657c9ee004b4fcc9f4bca/.travis.yml#L30

gettalong: That may be so but that line is from Travis and this is nothing that is generated by kramdown. As with .bundle, node_modules is something that the user decides, so it shouldn't be in the .gitignore file.

glebm: Done, though I disagree because this is the standard way of installing node dependencie.

gettalong:
Merged

PR #523, opened 2018-06-10
resolves #522 support HTML5 attribute values not in quotes

### LINKED ISSUE #522: HTML tag with attribute value not enclosed in quotes is not processed
In the HTML to native function, the parser is not recognizing (or processing) an HTML tag if it contains an attribute value that is not enclosed in quotes (e.g., ``. However, dropping the quotes around an attribute value is valid HTML.

Is this something you think Kramdown should permit? If so, I'd be happy to update the parser to support quote-less attributes.

mojavelinux:


gettalong:
Looks good to me! I have merged your changes and updated the syntax documentation.

mojavelinux:
Thanks @gettalong!

choldgraf:
Hey all - I'm curious if there is any plans at cutting a new release of Kramdown in the future? We've run into this bug for use w/ a Jekyll website (https://github.com/choldgraf/jupyter-book/issues/72) and would love to see it in a released version! No hurry, just trying to get an idea for what kind of timing to expect :-)

mojavelinux:
I would also be interested in a release, when time permits. Thanks!

gettalong:
@choldgraf @mojavelinux I'm in the process of cleaning the repo/moving stuff to the new kramdown org; release maybe this weekend, otherwise probably some time next week.

Issue #306, opened 2016-02-08
URLs are not autolinked in GFM mode


benbalter:
Per https://help.github.com/articles/autolinked-references-and-urls/#urls, in GFM, URLs should automatically become links, even when not wrapped in `<` and `>`.
## Steps to reproduce

`Link http://example.com`
## Expected

`Link <a href="http://example.com">http://example.com>
## Actual

`Link http://example.com`

gettalong:
Thanks for pointing this out but without the inside knowledge of what GFM **actually** supports I can't properly implement this.

benbalter:
> without the inside knowledge of what GFM actually supports I can't properly implement this.

Did a little digging, and it looks like [the autolink filter in HTML pipeline](https://github.com/jch/html-pipeline/blob/master/lib/html/pipeline/autolink_filter.rb), uses [Rinku](https://github.com/vmg/rinku), a Ruby gem which implements autolinking. It'd be an additional dependency, so its your call, but maybe make it an optional dependency that autolinks if present?

gettalong:
I would accept a pull request with Rinku being an optional dependency for kramdown. However, it should then be mentioned in kramdown's GFM docs that Rinku has to be installed for the GFM parser to work correctly.

gettalong:
Closed because feature request issues are now tracked through a dedicated project - see https://github.com/gettalong/kramdown/projects/1

zhengjia:
I attempted to fix this in https://github.com/gettalong/kramdown/pull/528. Hopefully someone can finish it off or work on a better fix.

cup:
https://github.com/gettalong/kramdown/issues/306
https://github.com/gettalong/kramdown/issues/490
https://github.com/gettalong/kramdown/pull/528

cup:
@gettalong why was this closed if it was not fixed?

gettalong:
@cup See [my comment above](https://github.com/gettalong/kramdown/issues/306#issuecomment-258595190)

cup:
I saw it before I posted mine. Can we please reopen the issue until it is actually fixed. Or is this a won't fix issue

gettalong:
If someone is actively working on a fix, we can re-open. All open pull requests/issues that need actions from contributers are tracked through the dedicated project.

cup:
[...]

Issue #490, opened 2018-01-25
Autolinking is not compatible with redcarpet and breaks GFM


abitrolly:
[kramdown autolinking](https://kramdown.gettalong.org/syntax.html#automatic-links) feature is strange. The better name for it is `semiautolinks`, because you still need to decorate every URL manually. This is different from [Redcarpet](https://github.com/vmg/redcarpet) `:autolink` extension and GFM autolinking https://help.github.com/articles/autolinked-references-and-urls/

gettalong:
Hmm... I don't really know what you want... kramdown's syntax is clearly defined by its syntax document. Redcarpet has a different syntax, so what? Each Markdown converter differs in some ways from most of the others. None is clearly better than the other (otherwise we would have a real Markdown standard).

If you like how redcarpet converts Markdown, please use Redcarpet. If you like how kramdown does its job, please use kramdown. But it is just a decision you have to make.

Aside from this, take a look at the [original Markdown spec](https://daringfireball.net/projects/markdown/syntax#autolink) by Gruber where you will find the reason why kramdown implements these autolinks.

Could you clarify what you want? And note that changes to kramdown's syntax have a very high chance of only being implemented in a new major version. Changes to kramdown's GFM parser are implemented if the current implementation doesn't match the GFM one and are considered bugs.

abitrolly:
Current implementation of GFM parser in kramdown doesn't match GFM - https://help.github.com/articles/autolinked-references-and-urls/#urls

kramdown GFM doesn't link URLs in pages that are linked by GitHub in source GFM. See https://kubernetes.io/docs/reference/generated/kubectl/kubectl/ for example.

gettalong:
Okay, then this has already been requested in #306 but there has not been a pull request for this. Since #306 is already tracked in the [Feature Requests Project](https://github.com/gettalong/kramdown/projects/1), I will close this issue.

abitrolly:
Ok. So #306 is not actually closed.. https://github.com/gettalong/kramdown/projects/1

gettalong:
It is closed because it is waiting for a pull request, as stated on the feature requests project page.

abitrolly:
That is strange. Usually issues are closed when they are resolved. This one is not resolved.

gettalong:
These issues are feature requests that need a pull request. So they are resolved from my point of view. And are therefore assigned to the feature requests project.

abitrolly:
[...]

PR #528, opened 2018-06-27
GFM autolink

zhengjia:
This implements the ability to parse a http/https link to html, e.g. `http://www.example.com`.
It fixes https://github.com/gettalong/kramdown/issues/306.

### DIFF
--- a/lib/kramdown/parser/gfm.rb
+++ b/lib/kramdown/parser/gfm.rb
@@ -7,7 +7,7 @@
 #++
 #
 
-
+require 'uri'
 require 'kramdown/parser'
 
 module Kramdown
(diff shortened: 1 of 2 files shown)


zhengjia:
CI failed. Looks like this should not be parsed to an html link (https://github.com/gettalong/kramdown/blob/master/test/testcases/block/09_html/not_parsed.text#L23)

```
<div>
http://example.com
</div>
```

zhengjia:
I ended up using a different regex and realized it's not easy to get the regex right and it might be different case-by-case. I'll close this PR but hopefully it might be useful for someone.

cup:
https://github.com/gettalong/kramdown/issues/306
https://github.com/gettalong/kramdown/issues/490
https://github.com/gettalong/kramdown/pull/528

PR #541, opened 2018-08-28
Run Travis tests with Bundler.

junaruga:
There are code to check `RUBY_VERSION` and ruby or jruby on Travis CI.
Using Bundler makes the logic simpler.

It is also good to reproduce running unit tests on developer's environment, as they can install the dependency gem packages under this project directory.

```
$ bundle install --path vendor/bundle
```

As a default behavior of `language: ruby`, `bundle install --jobs=3 --retry=3` is run in the install section on Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,15 +19,23 @@ matrix:
 
 before_install:
   - sudo apt-get update
-  - sudo apt-get install texlive-latex-base texlive-latex-recommended texlive-fonts-recommended texlive-latex-extra tidy
-  - gem install rake
-  - gem install minitest coderay stringex ritex execjs sskatex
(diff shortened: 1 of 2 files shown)


junaruga:
In the workflow, people have to install the dependency gems and `kramdown` gem to `~/.gem/ruby/2.5.0/gems/`. That is to invade the common space. People want to install those to project specific space?

Is it the only purpose that you do not want to use `Gemfile`?
You do not have to maintain the dependency gem information for both `Rakefile` and `Gemfile`?

gettalong:
Everyone can decide where they want to install things: some directory inside the project directory, using a ruby version manager like rbenv, rvm, chruby, or just using Ruby's default. If you want to use bundler, please go ahead and use it.

All the dependencies are listed in the gem specification, so it doesn't make sense to duplicate this information somewhere else.

PR #542, opened 2018-08-28
Update gitignore.

junaruga:
* Add "/" to ignore the top directories.
* Add ignored nvm install directory and katex directory that are created after installing the dependency packages.

For example, currently when creating below file `test/pkg/a.txt`, the file is ignored unintentionally.

```
$ mkdir test/pkg/
$ touch test/pkg/a.txt
```

But after this pull-request, it is appeared.

```
$ git status
On branch hotfix/gitignore
Untracked files:
  (use "git add <file>..." to include in what will be committed)

	test/pkg/
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,9 @@
-htmldoc
-pkg
-webgen-tmp
-CONTRIBUTERS
-VERSION
-kramdown.gemspec
-man/man1/kramdown.1
+/htmldoc
+/pkg
+/webgen-tmp
+/CONTRIBUTERS
+/VERSION
+/kramdown.gemspec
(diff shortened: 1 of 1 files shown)


gettalong:
I see what you get at but this is currently no problem and I don't think that it will be in the near future.

junaruga:
I see. How about the directories I added?

```
node_modules
katex
```

When you do tests on you local, you see the untracked files by `git status` don't you?

```
$ git status
On branch master
Your branch is up to date with 'origin/master'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	katex/
	node_modules/
```

PR #543, opened 2018-08-28
Add Travis build status badge on README.

junaruga:
Shall we add the Travis build status badge?
We can check the status easily.

I referred rubygems' case for the position of the badge. [1]

Here is the modified README by this pull-request.
You can check it.
https://github.com/junaruga/kramdown/blob/feature/ci-build-status-badge/README.md

[1] https://github.com/rubygems/rubygems/blob/master/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# kramdown
+# kramdown [![Build Status](https://travis-ci.org/gettalong/kramdown.svg?branch=master)](https://travis-ci.org/gettalong/kramdown)
 
 ## Readme first!
 


gettalong:
The `README.md` file of a project is a text file on purpose because it should be readable and understandable after you have cloned a project and when you view the README.md in a text editor. So an image link doesn't make much sense. Additionally, putting links to images which reside on external hosts into the README is not privacy-friendly.

Furthermore the badge wouldn't provide any additional information for a visitor since it would (nearly) always should that the build passes - if it doesn't the code is fixed in a short amount of time and it would pass again.

I don't deny that badges *might* be useful for projects that don't have a homepage but kramdown has a homepage with all the needed information on the front page.

The real solution for this is Github providing a way to configure badges that should be shown.

junaruga:
> The README.md file of a project is a text file on purpose because it should be readable and understandable 

> I don't deny that badges might be useful for projects that don't have a homepage but kramdown has a homepage with all the needed information on the front page.

Alright. The thoughts make sense. 
Thank you for the explanation.

PR #544, opened 2018-08-29
Remove duplicated bin-dir alias.

junaruga:
I just removed the duplicated bin-dir line.


gettalong:
I'm sorry but the `setup.rb` file should be left as is.

PR #551, opened 2018-10-08
Add html5 template with bootstrap 4.1

heitorPB:
Very simple page with the body wrapped in a bootstrap container.
The output page is a bit more friendly then the default `document` template.


gettalong:
Thank you for your pull request! Since this template represents a very specific use case, I don't think that it is useful to include it in kramdown itself.

PR #560, opened 2018-12-18
Add data attribute to code highlighted by Rouge

ashmaroli:
Given the following fenced-code-block:

    ```ruby
    "Hello World"
    ```

This proposal would result in the following markup:
```html
<div class="language-ruby highlighter-rouge">
  <div class="highlight">
    <pre class="highlight">
      <code data-lang="ruby">
        <span class="s2">"Hello World"</span>
      </code>
    </pre>
  </div>
</div>
```
## Use-case
I can then use the data attribute to render a **language label** adjacent to the code-block to inform the reader **what language that code block is about**.

I'll update the test scenario once you're on board with the change

### DIFF
--- a/lib/kramdown/converter/syntax_highlighter/rouge.rb
+++ b/lib/kramdown/converter/syntax_highlighter/rouge.rb
@@ -28,7 +28,9 @@ module Kramdown::Converter::SyntaxHighlighter
       return nil if opts[:disable] || !lexer
       opts[:css_class] ||= 'highlight' # For backward compatibility when using Rouge 2.0
       formatter = formatter_class(opts).new(opts)
-      formatter.format(lexer.lex(text))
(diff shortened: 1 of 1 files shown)


gettalong:
We already have the language available in the outer most `<div>` tag in the form `language-???`. There is no need to duplicate this information.

ashmaroli:
Is there any way that I could convince you to introduce an attribute that is solely the language name..? I mean just `="ruby"` instead of `="language-ruby highlighter-rouge"`
If not, is there an API available that would let me achieve a similar markup..?
Thanks

gettalong:
I get why you want the duplication but this is something very specific. We already have the `language-???` class which is [used by the spec](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-code-element).

If you really need this for your use case, please subclass the HTML converter and define your custom HTML output.

PR #564, opened 2019-01-09
Avoid allocating memory for unused block parameter

ashmaroli:
By using appropriate alternate methods or prefixing the unused parameter with an underscore (`_`)


gettalong:
Please provide benchmarks that show that these changes actually help performance/memory-wise.

PR #566, opened 2019-01-11
Convert if-else in append_text lambda to case-when

ashmaroli:
Then intention here is to improve readability especially with the nested if-else block when `child.type == :typographic_sym`.
The added advantage from this change is the reduced calls to `child.type` and `child.value` for each recursion.


gettalong:
Please move this pull request to the https://github.com/kramdown/parser-gfm repo - thank you!

PR #567, opened 2019-01-11
Improve inspect string for Kramdown::Element

ashmaroli:
Currently on `master`, inspecting the `@root` from `Kramdown::Parser::Base` outputs:
```
<kd:root nil {:encoding=>#<Encoding:UTF-8>, :location=>1, :options=>{}, :abbrev_defs=>{}, :abbrev_attr=>{}}>
```
The output string doesn't really help identifying the provided values.

With the proposed change, the same output  becomes:
```
<kd:root options={:encoding=>#<Encoding:UTF-8>, :location=>1, :options=>{}, :abbrev_defs=>{}, :abbrev_attr=>{}}>
```
and if `@root` was inspected later in the session, the output would become:
```
<kd:root options={:encoding=>#<Encoding:UTF-8>, :location=>1, :options=>{}, :abbrev_defs=>{}, :abbrev_attr=>{}} children=[<kd:text value="First line of a paragraph.\n" options={:location=>1}>]>
```

### DIFF
--- a/lib/kramdown/element.rb
+++ b/lib/kramdown/element.rb
@@ -501,7 +501,12 @@ module Kramdown
     end
 
     def inspect #:nodoc:
-      "<kd:#{@type}#{@value.nil? ? '' : ' ' + @value.inspect} #{@attr.inspect}#{options.empty? ? '' : ' ' + @options.inspect}#{@children.empty? ? '' : ' ' + @children.inspect}>"
+      str = +"<kd:#{@type}"
+      str << " value=#{@value.inspect}"       if !@value.nil?    && !@value.empty?
+      str << " attr=#{@attr.inspect}"         if !@attr.nil?     && !@attr.empty?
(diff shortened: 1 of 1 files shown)


gettalong:
I don't think that this change is really necessary, it just makes the output a bit larger. The attributes shown in the inspect string are self-explanatory, so no need for adding `options` or `children`.

ashmaroli:
> The attributes shown in the inspect string are self-explanatory, so no need for adding `options` or `children`.

How about atleast adding `attr=`? `@attr` is `nil` for `@root` (in the above example), yet not filtered out like the ivar `@value` is.

gettalong:
I will think about it.

gettalong:
@ashmaroli I have thought about this a bit more and will add all ivars to the inspect output (but with a single string interpolation instead of multiple operations).

PR #574, opened 2019-01-24
Add new CLI flag to output options from extensions

### LINKED ISSUE #569: Options from extracted gems are not output via CLI
After having installed `kramdown-2.0.0`, running `kramdown --help` doesn't surface `options` from the extracted gems.

```bash
kramdown --help
```

I also used a `Gemfile` and tried again with `bundle exec`. Still no success!
```ruby
# Gemfile

gem 'kramdown', '~> 2.0'
gem 'kramdown-parser-gfm'
```
```bash
$ bundle exec kramdown --help
```

What should a user do to surface `options` from `kramdown-parser-gem` gem in the `--help` output?

ashmaroli:
Resolves #569 

## Summary
- `"-x", "--extend-help ARG"`
- Loads given extension gem and displays all available kramdown options
- Outputs error message and aborts if `LoadError` is raised.
- defined helper methods to avoid repetition.

### DIFF
--- a/bin/kramdown
+++ b/bin/kramdown
@@ -30,6 +30,55 @@ end
 options = {}
 format = ['html']
 
+def load_extensions(gems)
+  gems.each do |extension|
+    gem_name = "kramdown-#{extension}"
+    begin
+      require gem_name
+    rescue LoadError
+      puts <<-MSG
+
+        EXTENSION NOT FOUND:  #{gem_name}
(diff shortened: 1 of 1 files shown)


gettalong:
Thanks for the pull request but it only allows displaying the options, without being able to actually use them. I have committed a change that solves the problem.

PR #575, opened 2019-01-28
Add math_engine without iteration

ashmaroli:
This removes some redundancy with adding the lone `math_engine` in kramdown core: `Mathjax`

## Motivations
- It appears wasteful to iterate through a ***known** single-item array* and then referring to that single-item in the associated block. (allocation for array, allocation for block parameter)
- Unnecessary allocation for a local variable to hold a Symbol with a known name.
- Unnecessary interpolations for a known string.

## Advantages
- Addresses the motivation listed above
- Absence of interpolation allows the resulting string to be frozen automatically.
- Absence of interpolation allows removing the now redundant use of `Kramdown::Utils.deep_const_get()`

## Disadvantages
- Makes addition of new `math_engine` to *kramdown core* harder.

### DIFF
--- a/lib/kramdown/converter.rb
+++ b/lib/kramdown/converter.rb
@@ -49,18 +49,15 @@ module Kramdown
 
     configurable(:math_engine)
 
-    ["Mathjax"].each do |klass_name|
-      kn_down = klass_name.downcase.intern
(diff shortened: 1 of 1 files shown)


gettalong:
Thanks but it was a deliberate decision to leave it this way for the time being.

PR #578, opened 2019-02-01
Ask an element if it is a :block or :span directly

ashmaroli:
Allows simplifying calls such as `Kramdown::Element.category(el) == :block` or
`Kramdown::Element.category(el) == :span` with **`el.block?`** and **`el.span?`** respectively.


gettalong:
@ashmaroli Thank you for this contribution! I have rebased and merged your changes.

PR #585, opened 2019-03-29
use travis apt addon

stephengroat:


gettalong:
What would be the benefit?

stephengroat:
Removing the requirement for `sudo` allows the job to use the container based infrastructure, which is faster to boot, etc.

gettalong:
Thanks for the explanation! I have merged your changes.

PR #586, opened 2019-04-01
Fix YARD warning UnknownTag

olleolleolle:
This PR avoids a warning "UnknownTag".


gettalong:
kramdown uses RDoc for its API documentation, not YARD. So there is nothing to fix.

PR #590, opened 2019-04-14
syntax.page: Use `aligned` instead of `align*`

glebm:
`aligned` should be used instead of `align*` in math mode.

While MathJax supports both in math mode, it isn't standard and is not supported in other engines, such as KaTeX and LaTeX.


gettalong:

PR #597, opened 2019-05-06
Use static regex for simple codespan results

ashmaroli:
## Rationale

It is wasteful to allocate <code>/\`/</code> and <code>/\`\Z/</code> every time a simple codespan is parsed. Instead use a static regex which is allocated only once each.

## Context
Results from profiling Jekyll docs:

### Before
```
Total allocated: 158.07 MB (1168216 objects)
```
### After
```
Total allocated: 157.1 MB (1162835 objects)
```

### DIFF
--- a/lib/kramdown/parser/kramdown/codespan.rb
+++ b/lib/kramdown/parser/kramdown/codespan.rb
@@ -27,16 +27,10 @@ module Kramdown
 
         # assign static regex to avoid allocating the same on every instance
         # where +result+ equals a single-backtick. Interpolate otherwise.
-        if result == '`'
-          scan_pattern = /`/
-          str_sub_pattern = /`\Z/
-        else
-          scan_pattern = /#{result}/
-          str_sub_pattern = /#{result}\Z/
-        end
+        scan_pattern = (result == '`') ? /`/ : /#{result}/
 
         if (text = @src.scan_until(scan_pattern))
-          text.sub!(str_sub_pattern, '')
+          text.chomp!(result)
           unless simple
             text = text[1..-1] if text[0..0] == ' '
             text = text[0..-2] if text[-1..-1] == ' '
--- a/lib/kramdown/parser/kramdown/codespan.rb
+++ b/lib/kramdown/parser/kramdown/codespan.rb
@@ -25,8 +25,18 @@ module Kramdown
           return
         end
 
-        if (text = @src.scan_until(/#{result}/))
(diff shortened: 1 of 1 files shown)


ashmaroli:
### Profiler Diff Summary
```diff
--- master branch https://travis-ci.org/gettalong/kramdown/jobs/536192573
+++ PR branch     https://travis-ci.org/gettalong/kramdown/jobs/536323986

- Total allocated: 220.89 MB (1783975 objects)
+ Total allocated: 219.34 MB (1775370 objects)
  Total retained:  1.11 MB (1582 objects)
```

lib/kramdown/parser/kramdown/codespan.rb
```diff
@@ -25,8 +25,18 @@ def parse_codespan
...
+        if result == '`'
+          scan_pattern = /`/
+          str_sub_pattern = /`\Z/
+        else
+          scan_pattern = /#{result}/
+          str_sub_pattern = /#{result}\Z/
+        end
+
+        if (text = @src.scan_until(scan_pattern))
+          text.sub!(str_sub_pattern, '')
```
gettalong: At least for this line we can do even better by using `text.chomp!(result)` instead of `#sub!`, and don't need a regexp at all. Then the `if` clause can be inlined as ternary operator in the line above this one.

ashmaroli: That's a splendid suggestion, Thomas! :clap: :clap:

ashmaroli:
@gettalong I guess we were wrong. The latest change isn't synonymous with the existing behavior:
```
irb(main):001:0> string = "`foobar`\n"
=> "`foobar`\n"

irb(main):002:0> string.sub(/`\Z/, '')
=> "`foobar\n"

irb(main):003:0> string.chomp('`')
=> "`foobar`\n"

irb(main):004:0> string.sub(/`\Z/, '') == string.chomp('`')
=> false
```
Shall I revert the latest commit or am I testing incorrectly?

gettalong:
You are right, `\Z` matches before a new line at the end, `\z` matches at the end. I guess the original version it is.

ashmaroli:
@gettalong Pinging you as a gentle reminder...

gettalong:
@ashmaroli Sorry for the delay, I will find some time in the coming days for kramdown.

gettalong:
@ashmaroli Even though the winnings are not much, the changes are not very invasive and I will therefore accept them.

PR #601, opened 2019-05-15
Interpolate stateless regexps just once

ashmaroli:
If an interpolated regular expression within a method definition and if the interpolation doesn't depend on the state of the instance, then they should essentially be **interpolated just once** to avoid multiple allocations and every call to the associated method.


gettalong:
Thanks for your contribution - I have merged the changes!

PR #603, opened 2019-05-15
Add rake task to profile memory usage of testcases

ashmaroli:
This RakeTask runs assertions on the `testcases/` files, profiles memory usage and prints out the memory usage stats (to a file locally and to `stdout` on CI)
(*Additionally, Travis CI has been configured to run this task should you decide to re-enable it for this repo.*)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,14 @@ rvm:
   - 2.3
   - 2.4
   - 2.5
-  - 2.6
+  - &latest_ruby 2.6
   - jruby-head
 
 matrix:
+  include:
+    - rvm: *latest_ruby
+      script: rake dev:profile_memory
+      name: Profile Memory Usage
   allow_failures:
     - rvm: jruby-head
 
@@ -20,11 +24,9 @@ addons:
       - texlive-latex-extra
 
 before_install:
-  - gem install rake
-  - gem install minitest stringex
-  - gem install rouge
   - wget https://github.com/htacg/tidy-html5/releases/download/5.4.0/tidy-5.4.0-64bit.deb
   - sudo apt purge libtidy-0.99-0
   - sudo dpkg -i tidy-5.4.0-64bit.deb
 
+install: gem install memory_profiler minitest rake rouge stringex
 script: ruby -Ilib:test test/test_*
--- a/Rakefile
+++ b/Rakefile
@@ -215,6 +215,12 @@ namespace :dev do
     end
   end
 
+  desc "Profile memory usage from the test-cases"
+  task :profile_memory do
+    puts ""
+    ruby 'benchmark/profile_memory_usage.rb'
+  end
+
   desc "Upload the website"
   task publish_website: ['doc'] do
     puts "Transfer manually!!!"
--- a/benchmark/profile_memory_usage.rb
+++ b/benchmark/profile_memory_usage.rb
@@ -1,0 +1,78 @@
+# frozen_string_literal: true
+
+$LOAD_PATH.unshift File.join(__dir__, '../lib')
+
+require 'kramdown'
+require 'memory_profiler'
+require 'minitest'
+require 'rouge'
+
+class Rouge::Formatters::HTMLLegacy
+  def format(tokens, &b)
+    super(tokens, &b).sub(/<\/code><\/pre>\n?/, "</code></pre>\n")
+  end
+end
+
+class RougeHTMLFormatters < Rouge::Formatters::HTMLLegacy
+  def stream(tokens, &b)
+    yield '<div class="custom-class">'
+    super
+    yield '</div>'
+  end
+end
+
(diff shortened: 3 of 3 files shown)


.gitignore
```diff
@@ -3,5 +3,7 @@ pkg
 webgen-tmp
 CONTRIBUTERS
 VERSION
+Gemfile
+Gemfile.lock
```
gettalong: This project doesn't use bundler, so these two lines are not required.

.travis.yml
```diff
@@ -6,9 +6,6 @@ rvm:
   - 2.6
   - jruby-head
 
-jdk:
-  - oraclejdk8
-
```
gettalong: Why the removal?

.travis.yml
```diff
@@ -24,7 +21,9 @@ addons:
 
 before_install:
   - gem install rake
-  - gem install minitest stringex
-  - gem install rouge
+  - rake dev:gemspec
+  - rake dev:gemfile
 
-script: ruby -Ilib:test test/test_*
+script: bundle exec rake dev:profile_memory
```
gettalong: This project doesn't use bundler since it is not necessary. And Travis should run the tests, not profile memory (even if that may run some tests).

.travis.yml
```diff
@@ -24,7 +21,9 @@ addons:
...
+notifications:
+  email: false
```
gettalong: Notifications should be enabled.

Rakefile
```diff
@@ -215,6 +215,25 @@ namespace :dev do
...
+      gem 'memory_profiler'
+      gem 'rake'
+    TEXT
+  end
+
+  desc "Profile memory usage from the test-cases"
+  task profile_memory: [:gemspec, :gemfile] do
+    puts ""
+    ruby 'test/profile_memory_usage.rb'
+  end
```
gettalong: Why would you need the `gemfile`? We only need to make sure that the `memory_profiler` and `colorator` gems are installed for this to work, a simple hint to stdout that one would need to run `gem install colorator memory_profile` in case of errors would be sufficient.

test/profile_memory_usage.rb
```diff
@@ -0,0 +1,78 @@
+# frozen_string_literal: true
+
+require 'bundler/setup'
```
gettalong: Please don't rely on bundler.

test/profile_memory_usage.rb
```diff
@@ -0,0 +1,78 @@
+# frozen_string_literal: true
```
gettalong: This file should probably go into the `benchmark` directory since it is benchmarking the library?

.travis.yml
ashmaroli: The Travis builds on my fork failed as it couldn't install the package:
https://travis-ci.org/ashmaroli/kramdown/jobs/532961008

.travis.yml
ashmaroli: > And Travis should run the tests, not profile memory (even if that may run some tests).

:smiley: I figured since Travis wasn't running on the project currently, you didn't want automated testing.. :stuck_out_tongue_winking_eye:

test/profile_memory_usage.rb
ashmaroli: I put it here since it runs assertions on the `testcases` as well. But yes, you're right. The primary intention of the script is profiling after all.....

.travis.yml
gettalong: Travis is running but tidy has a problem with one test... Locally everything is fine, so I don't really know what Travis is up to. I have updated the code to produce the error messages from tidy, will see if that helps.

.travis.yml
gettalong: I have update the whole `.travis.yml` file and it seems to work fine now...

gettalong:
I have looked at the profiling file and was wondering whether a different approach would be more useful.

I have created several scripts for HexaPDF which allow easy profiling/benchmarking of arbitrary code. In case of the `memory_profiler` gem it looks like this:

~~~ ruby
BEGIN {
  require 'memory_profiler'
  MemoryProfiler.start
}

END {
  report = MemoryProfiler.stop
  report.pretty_print
}
~~~

By requiring this file it is possible to use it for any Ruby executable like this:

~~~shell
$ ruby -Ilib:test:benchmark -r profile_memory test/test_files.rb
~~~

The advantage would be that it could be run on several different executables and that those wouldn't need to be changed.

What do you think?

ashmaroli:
> What do you think?

SGTM! My primary intention was solely to gauge memory allocations from a given Pull Request via CI.

Feel free to either cherry-pick the commit from this branch and proceed, or push an entirely fresh commit directly to the `master`..

gettalong:
Thanks for your work - I have pushed a commit with these changes.

PR #606, opened 2019-05-23
Fix formatting in syntax documentation

tom93:
Put the example that shows how quotes can be escaped with backslashes
inside a code span to display the backslashes.


gettalong:

PR #607, opened 2019-05-24
Allow running dev:profile_memory task on Windows

ashmaroli:
PATH_SEPARATOR on Windows is a semicolon (`;`) instead of the colon (`:`)
Therefore, using `File::PATH_SEPARATOR` constant allows running the task
across multiple platforms


gettalong:

PR #612, opened 2019-06-09
Introduce Kramdown::Registry to stash doc options

### LINKED ISSUE #610: Initializing multiple Documents with the same options
When given the following code:
```ruby
('A'..'Z').to_a.each do |name|
  (1..50).to_a.each do |section|
  source = File.read("folio/#{name}#{section}", mode: 'utf-8')
  Kramdown::Document.new(source, FOLIO_OPTIONS).to_html
end
```
Each of the `Document` instance generates a copy of the same `Options` hash (same `#hash` value but different `#object_id`). This involves consuming resources to parse and merge hashes at multiple levels to produce *the same output* from *the same input*.

Would be great if this could be optimized.

ashmaroli:
This is a preliminary attempt to resolve #610 
@gettalong If you feel, this route is acceptable, I will try and add tests for the API

### DIFF
--- a/lib/kramdown/converter/base.rb
+++ b/lib/kramdown/converter/base.rb
@@ -99,7 +99,9 @@ module Kramdown
       # 4. Check if the template name starts with 'string://' and if so, strip this prefix away and
       #    use the rest as template.
       def self.convert(tree, options = {})
-        converter = new(tree, ::Kramdown::Options.merge(options.merge(tree.options[:options] || {})))
+        opts = options.merge(tree.options[:options] || {})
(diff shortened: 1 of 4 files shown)


gettalong:
@ashmaroli I'm not sure whether this is a good way forward. What if a parser/converter modifies the options? Then all further invocations would use wrong data. So we would at least need to return a `#dup`ped hash. This would negate the memory savings but not the compute savings. Also `OptionsCache` seems to be a more appropriate class name.

Have you done a benchmark of time/memory usage with and without your changes? Is there a significant difference?

ashmaroli:
Perhaps the added complexity will hamper maintainability down the road as well..
Let us nip it in the bud then..
Thanks for looking into this anyways.

PR #614, opened 2019-06-24
Fix a bug that kept us from adding the markdown attribute to HTML tags that contain valid markdown when parsing HTML

boblail:
You can reproduce the bug like this:

> ```ruby
> html = <<~HTML
> <div>
>   alt
> </div>
> HTML
> 
> markdown = Kramdown::Document.new(html, input: :html).to_kramdown
> # => "<div>\nalt\n</div>"
> 
> html = Kramdown::Document.new(markdown).to_html
> # => "<div>\nalt\n</div>\n\n"
> ```

The bug is that the call to `to_kramdown` converted the IMG tag but **_did not_** add `markdown="1"` to the DIV tag.

As of Kramdown 2.1.0, the `markdown` attribute is added if the IMG tag is wrapped in a paragraph; and you can see that when the `markdown` attribute is present, the HTML roundtrips successfully:

> ```ruby
> html = <<~HTML
> <div>
>   <p>alt</p>
> </div>
> HTML
> 
> markdown = Kramdown::Document.new(html, input: :html).to_kramdown
> # => "<div markdown=\"1\">\nalt\n</div>"
> 
> html = Kramdown::Document.new(markdown).to_html
> # => "<div>\n  <p></p>\n\n</div>"
> ```

### DIFF
--- a/lib/kramdown/converter/kramdown.rb
+++ b/lib/kramdown/converter/kramdown.rb
@@ -184,10 +184,14 @@ module Kramdown
 
       HTML_TAGS_WITH_BODY = ['div', 'script', 'iframe', 'textarea']
 
+      NON_MARKDOWN_TYPES = [:entity, :text, :html_element].freeze
+      private_constant :NON_MARKDOWN_TYPES
+
       def convert_html_element(el, opts)
         markdown_attr = el.options[:category] == :block && el.children.any? do |c|
-          c.type != :html_element && (c.type != :p || !c.options[:transparent]) &&
-            Element.category(c) == :block
+          c.type != :html_element &&
+          !(c.type == :p && c.options[:transparent] && c.children.none? { |t| !NON_MARKDOWN_TYPES.member?(t.type) }) &&
+          Element.category(c) == :block
         end
         opts[:force_raw_text] = true if %w[script pre code].include?(el.value)
         opts[:raw_text] = opts[:force_raw_text] || opts[:block_raw_text] || \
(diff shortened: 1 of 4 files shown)


lib/kramdown/converter/kramdown.rb
```diff
@@ -186,8 +186,9 @@ def convert_dt(el, opts)
 
       def convert_html_element(el, opts)
         markdown_attr = el.options[:category] == :block && el.children.any? do |c|
-          c.type != :html_element && (c.type != :p || !c.options[:transparent]) &&
-            Element.category(c) == :block
+          c.type != :html_element &&
+          !(c.type == :p && c.options[:transparent] && c.children.none? { |t| ![:entity, :text, :html_element].member?(t.type) }) &&
```
ashmaroli: `[:entity, :text, :html_element]` is a static array. In order to avoid repeated allocations of this array on multiple iterations, I recommend hoisting the array as a private constant:
```ruby
TYPES_ARRAY = [:entity, :text, :html_element]
private_constant :TYPES_ARRAY

def convert_html_element(el, opts)
  markdown_attr = el.options[:category] == :block && el.children.any? do |c|
    ...
    c.children.none? { |t| !TYPES_ARRAY.include?(t.type) }
    ...
  end
end
```

gettalong: How did you determine these three element types?

ghepting:
Nice bug fix @boblail! :)

lib/kramdown/converter/kramdown.rb
boblail: @gettalong, I added element types until the rest of the Kramdown spec suite passed without changes. 😄 

* * *

I considered 

```ruby
c.children.none? { |t| MARKDOWN_TYPES.member?(t.type) }
```

instead of 

```ruby
c.children.none? { |t| !NON_MARKDOWN_TYPES.member?(t.type) }
```

but I think the list would be longer.

boblail:
@gettalong, is this OK to merge? Or did you have any followup questions?

gettalong:
Thank you for your contribution - I have cherry-picked your commit and adapted it a bit.

PR #619, opened 2019-07-14
Reduce string allocations from adapting source

ashmaroli:
`String#encode` returns a copy of the given `source` string.
Subsequent operations on this copy can be made in place to avoid allocating intermediate strings.


gettalong:
Thanks!

PR #620, opened 2019-07-14
Use String#capitalize! to avoid interim allocation

ashmaroli:
Using `String#[]` results in allocating the substring which is again duplicated via `String#upcase`.
In effect, *camelizing* `"foo_bar"` originally involved allocating `"f"`, `"F"`, `"oo"`, `"b"`, `"B"`, `"ar"` unnecessarily.

### DIFF
--- a/lib/kramdown/utils.rb
+++ b/lib/kramdown/utils.rb
@@ -24,7 +24,7 @@ module Kramdown
 
     # Treat +name+ as if it were snake cased (e.g. snake_case) and camelize it (e.g. SnakeCase).
     def self.camelize(name)
-      name.split('_').inject(+'') {|s, x| s << x[0..0].upcase << x[1..-1] }
+      name.split('_').inject(+'') {|s, x| s << x.capitalize! }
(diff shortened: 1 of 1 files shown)


gettalong:
This change is not backwards-compatible:

~~~
2.5.0 :001 > "ASDF".capitalize!
 => "Asdf"
2.5.0 :002 > "asdf".capitalize!
 => "Asdf"
~~~

`String#capitalize!` not only uppercases the first character but also lowercases all others; and this is a problem for some at least some parsers/converters (e.g. `HashAST`).

ashmaroli:
Oh! That's unfortunate.
Looks like the `HashAST` converter isn't covered by the test-suite. Travis should've failed the build otherwise..

gettalong:
The test suite consists mostly of sample files and the expected output. There are not many unit tests.

ashmaroli:
I understand that. What I meant was that the test-suite doesn't test the expected output from the `HashAST` converter.

PR #621, opened 2019-07-14
Stash array of numbers in a private constant

ashmaroli:
This is to reduce allocation of `(0..128).to_a` in applicable situations.

`2/3` of the conditional in `#convert_ul` in the HTML converter invokes a new array `(0..128).to_a` which is subsequently duplicated via `Array#map`. This can be optimized by instead using a constant array of integers which are then duped as required.


gettalong:
Thanks!

PR #629, opened 2019-10-11
s/a HTML/an HTML/ in documents

284km:
I found a typo in syntax page.
https://kramdown.gettalong.org/syntax.html

I fixed it and all same others.


gettalong:
Thanks!

Issue #211, opened 2015-01-20
Support starting numbered lists with arbitrary number


winniehell:
Allow numbered lists to have a defined offset like [in CommonMark](http://spec.commonmark.org/dingus.html?text=5.%20one%0A1.%20two).

Current output:

```
$ kramdown --output kramdown <<< '5.  five
6.  six'

1.  five
2.  six
```

Expected:

```
$ kramdown --output kramdown <<< '5.  five
6.  six'

5.  five
6.  six
```

Especially useful in quotes:

```
$ kramdown --output kramdown <<< '> 42.  Question

What is it?'
```

should really be

```
> 42.  Question

What is it?
```

gettalong:
Thanks for the issue. However, since this will change the output and probably break some documents, this can't be done in the 1.x release and has to wait for the 2.x release.

winniehell:
@gettalong 2.x would be great :+1:

Phrogz:
+1 particularly for inline quoted responses. When someone writes: 

    Some questions:

    1. How big are your shoes?
    2. Can a bullet ricochet off an armadillo?

And I respond with a quoting system that prefixes the input with `>`, and inject my responses:

    > 1. how big are your shoes
    
    Normal sized

    > 2. Can a bullet ricochet off an armadillo?

    [Apparently, yes.](http://www.bbc.com/news/world-us-canada-33748027)

The result is undesirable.

> 1. how big are your shoes
   
Normal sized

> 2. Can a bullet ricochet off an armadillo?

[Apparently, yes.](http://www.bbc.com/news/world-us-canada-33748027)

My workaround hack is to wrap the quoted list items in backticks to preserve the typed value.

e9t:
:+1:

chimericdream:
Another option that I just found is to use an IAL declaration before the list. I was looking for this exact feature and was sad (at first) to learn that it isn't supported. However, the following works as desired:

```
{:start="3"}
1. test
1. test
1. test
```

Output:
```
3. test
4. test
5. test
```

Tested with this online Kramdown editor: https://kramdown.herokuapp.com/

This doesn't solve the problem for the average user who doesn't know much markdown, but it's a workaround for people using Kramdown in cases like Jekyll sites and whatnot.

winniehell:
@chimericdream That is awesome, thank you very much! :+1: 

However for quoting multiple list entries, this is a lot of overhead:

```markdown
> {:start="42"}     
> 42. Question

Nobody knows!

[...]

PR #631, opened 2019-10-12
Add "start" attribute to <ol> for HTML conversion

ken-mu:
## Background

I can see the difference between GitHub README and GitHub Pages like following ( Please see contents of Header5)

- README: https://github.com/ken-mu/cayman/blob/master/index.md
- GitHub Pages: https://ken-mu.github.io/cayman/

README lists ordered list (1. 2. 3.), but GitHub Pages lists only 1. at all.

## Changes

This change supports "start" attribution for <ol> HTML tag.
markdown:

```
# basic ol

1. test1
2. test2

# separated ol

1. test1

test

2. test2
3. test3

# ul

- test1
- test2
```

HTML:

```
<h1 id="basic-ol">basic ol</h1>

<ol>
  <li>test1</li>
  <li>test2</li>
</ol>

<h1 id="separated-ol">separated ol</h1>

<ol>
  <li>test1</li>
</ol>

<p>test</p>

<ol start="2">
  <li>test2</li>
  <li>test3</li>
</ol>

<h1 id="ul">ul</h1>

<ul>
  <li>test1</li>
  <li>test2</li>
</ul>
```

## Tests

I'll add testcases if necessary.

### DIFF
--- a/lib/kramdown/parser/kramdown/list.rb
+++ b/lib/kramdown/parser/kramdown/list.rb
@@ -47,6 +47,7 @@ module Kramdown
       LIST_START_UL = /^(#{OPT_SPACE}[+*-])([\t| ].*?\n)/
       LIST_START_OL = /^(#{OPT_SPACE}\d+\.)([\t| ].*?\n)/
       LIST_START = /#{LIST_START_UL}|#{LIST_START_OL}/
+      LIST_START_NUM_OL = /^#{OPT_SPACE}(\d)+\./
 
       # Parse the ordered or unordered list at the current location.
       def parse_list
@@ -59,6 +60,7 @@ module Kramdown
         eob_found = false
         nested_list_found = false
         last_is_blank = false
+	list_first = true
         until @src.eos?
           start_line_number = @src.current_line_number
           if last_is_blank && @src.check(HR_START)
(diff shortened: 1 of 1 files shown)


gettalong:
Please see https://github.com/gettalong/kramdown/issues/211 for why this change can only be implemented for the next major version.

A working solution is to use an IAL:

~~~
{: start="3"}
1. item 3
2. item 4
~~~

ken-mu:
@gettalong Thank you for information.
Right, I confirmed it has been planned for changing someday...

About IAL, that would be solution at once.
But I will not use IAL because that is not portable. 
Some other Markdown site (GitHub README, CommonMarkGhPages) don't support IAL. 
I cannot use kramdown specific features because we should standardize our code as common Markdown notation.

I'm very happy if you can resolve #211 as general language like this PR.

gettalong:
I recommend staying with just *one* Markdown parser/converter since the dialects are nearly always not compatible with one another and there is **no standard Markdown syntax** (even the implementation of the basic features of the original Markdown syntax differs from implementation to implementation...).

So: If you decide to use kramdown, use the syntax kramdown provides, including IALs.

marelons1337:
Hi,
Was that picked up after this issue was closed? As of 2.4.0 it doesn't seem to work and ordered lists always start with 1. with GFM, and IAL doesn't seem to work for it either.

gettalong:
@marelons1337 The solution in https://github.com/gettalong/kramdown/pull/631#issuecomment-541356100 works. The output is

~~~ html
<ol start="3">
  <li>item 3</li>
  <li>item 4</li>
</ol>
~~~

marelons1337:
Hi Thomas, thanks for looking at this, comes out I didn't have `start` attribute whitelisted, so in my sanitizer.
Is there any chance to include GFM-like syntax where it actually starts the list from the number provided as the first one, or should I just try adding it to https://github.com/kramdown/parser-gfm?

gettalong:
Changing this would be a breaking change, potentially modifying the result of running kramdown on existing documents. Therefore this won't be done in a minor release. And I also currently don't have any intentions of implementing this in a major release.

PR #638, opened 2020-01-14
Add missing rexml dependency

deivid-rodriguez:
The next ruby version won't have rexml as a default gem, so it will need to be explicitly added as a dependency.

See https://bugs.ruby-lang.org/issues/16485#change-83794.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -173,6 +173,7 @@ namespace :dev do
       s.executables = ['kramdown']
       s.default_executable = 'kramdown'
       s.required_ruby_version = '>= 2.3'
+      s.add_dependency "rexml"
       s.add_development_dependency 'minitest', '~> 5.0'
       s.add_development_dependency 'rouge'
       s.add_development_dependency 'stringex', '~> 1.5.1'


ashmaroli:
> The next ruby version won't have rexml as a default gem

*Just curious..*

Isn't the next ruby version Ruby 3.0?
In that case wouldn't it be better to have all Ruby 3.0 compatibility changes in one PR..? Or does Ruby 2.7 throw deprecation warnings regarding rexml as well?

deivid-rodriguez:
I test my library against ruby-head to be able to early address any ruby-core incompatibilities. My build against ruby-head broke because of the mentioned change, since under current ruby-head `kramdown` can no longer be required unless you manually depend on `rexml`.

This change should be correct regardless, because `rexml` is a gem, and `kramdown` depends on it. That said, I understand if you don't want to fix this this early, and I can easily workaround this by adding `rexml` to my `Gemfile` to account for `kramdown` lacking the dependency untill `kramdown` catches up.

Just let me know and I'll close the PR accordingly.

ashmaroli:
@deivid-rodriguez I was just asking out of curiosity. I'm not a maintainer of this project..

deivid-rodriguez:
Oh, ok, sorry. Well, I hope I answered your question in any case :)

deivid-rodriguez:
I should probably also add that this change _does_ have an impact. For example, if `rexml` makes a disastrous release that breaks everything, `kramdown` would be affected. But on the other hand, if `rexml` fixes a security vulnerability, `kramdown` will automatically pick that up :)

deivid-rodriguez:
Actually, those same things can already happen currently, but the changes only happen accross ruby releases. With this PR, any change in the `rexml` dependency becomes more "agile" because it only requires a gem release, not a release of the full ruby language.

gettalong:
@deivid-rodriguez Thank you for your contribution!

deivid-rodriguez:
Thanks for merging! :heart:

caser:
I'm still getting this error:

`/.rvm/gems/ruby-head/gems/kramdown-1.17.0/lib/kramdown/parser/html.rb:10:in `require': cannot load such file -- rexml/parsers/baseparser (LoadError)`

Is this related to above?

deivid-rodriguez:
I don't think this fix has been released yet, so could be related, yeah. Does it get fixed if you manually install `rexml` or add it to your `Gemfile`?

gettalong:
@caser Please try the latest version of kramdown and see if it works!

JetForMe:
Any progress on this? I just ran into it, as kramdown is a transitive dependency. I had to explicitly include rexml, which now makes my install a little more fragile.

ashmaroli:
@JetForMe This change was shipped in `kramdown-2.2.0`. So, you may have to update your installed version.

PR #641, opened 2020-02-20
Use private constants to define :toc_levels option

ashmaroli:
## Motivation
*This is to reduce array allocation when multiple `Kramdown::Document` objects are initialized with the same `:toc_levels` value.*

## Summary

When `Kramdown::Document` is initialized with option `:toc_levels => [1, 2, 3, 4, 5, 6]` use a predefined constant with identical contents instead of generating a new array from the option value or when `Kramdown::Document` is initialized with option `:toc_levels => (1..6)`, use a predefined constant array instead of converting the range into a new Array.

*It is safe to use constant values because the (merged) `options` attribute of `Kramdown::Document` instances is already frozen on initialization.*

### DIFF
--- a/lib/kramdown/options.rb
+++ b/lib/kramdown/options.rb
@@ -328,7 +328,11 @@ module Kramdown
       Used by: HTML converter, kramdown converter
     EOF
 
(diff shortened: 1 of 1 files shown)


gettalong:
Thank you!

PR #642, opened 2020-02-20
Reduce conversion of Symbol to String

ashmaroli:
Instead of interpolating `:rouge` (default and therefore more common scenario) to `"rouge"` multiple times when a `Kramdown::Document` is converted `to_html`, use a *frozen_string* to be used per instance.

### DIFF
--- a/lib/kramdown/converter/html.rb
+++ b/lib/kramdown/converter/html.rb
@@ -47,13 +47,8 @@ module Kramdown
         @indent = 2
         @stack = []
 
-        @highlighter = options[:syntax_highlighter]
-        if @highlighter == :rouge
-          # assign a frozen string instead of converting Symbol to String for default use-case.
-          @highlighter = "rouge"
-        else
-          @highlighter = @highlighter.to_s
-        end
+        # stash string representation of symbol to avoid allocations from multiple interpolations.
+        @highlighter = options[:syntax_highlighter].to_s
       end
 
       # The mapping of element type to conversion method.
--- a/lib/kramdown/converter/html.rb
+++ b/lib/kramdown/converter/html.rb
@@ -52,7 +52,7 @@ module Kramdown
           # assign a frozen string instead of converting Symbol to String for default use-case.
           @highlighter = "rouge"
         else
-          @highlighter = options[:syntax_highlighter].to_s
+          @highlighter = @highlighter.to_s
         end
       end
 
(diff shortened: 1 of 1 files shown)


lib/kramdown/converter/html.rb
```diff
@@ -46,6 +46,14 @@ def initialize(root, options)
...
         @indent = 2
         @stack = []
+
+        @highlighter = options[:syntax_highlighter]
+        if @highlighter == :rouge
+          # assign a frozen string instead of converting Symbol to String for default use-case.
+          @highlighter = "rouge"
+        else
+          @highlighter = @highlighter.to_s
+        end
```
gettalong, line 56: Even if rouge *is* the default, I don't like hard-coding it this way.

Storing the stringified syntax highlighter is probably only useful when *many* code/pre sections exist. Otherwise I think the memory/cpu usage is negligible.

lib/kramdown/converter/html.rb
```diff
@@ -162,24 +170,26 @@ def convert_ul(el, indent)
         elsif !@footnote_location && el.options[:ial] && (el.options[:ial][:refs] || []).include?('footnotes')
           @footnote_location = ZERO_TO_ONETWENTYEIGHT.map { rand(36).to_s(36) }.join
         else
-          format_as_indented_block_html(el.type, el.attr, inner(el, indent), indent)
+          type = el.type == :ul ? "ul" : "ol"
+          format_as_indented_block_html(type, el.attr, inner(el, indent), indent)
```
gettalong, line 174: Moving to a string instead of `el.type` is okay if it is a direct exchange. But if we have to check beforehand, even with only two choices, I think that this unnecessary complicates the code. So please remove this part and the others where additional checks are needed.

lib/kramdown/converter/html.rb
ashmaroli: *many* here is subjective..
I just ran this branch against the latest commit on `jekyll/jekyll`. Its docs site returned a total of `2894` calls to the `Converter::Html#add_syntax_highlighter_to_class_attr` method.
(*each call would convert `:rouge` to `"rouge"`*)

Is that *many* or just *not that many*..?

gettalong: Hmm... I think the way it is now is a good compromise and I hope you are okay with it. There are also [discussions](https://bugs.ruby-lang.org/issues/16153) going on whether `Symbol#to_s` should return a frozen string in the future.

ashmaroli: Yes, I'm okay with the *current* diff. :)

gettalong:
Perfect, thank you! I have squash-merged your changes and additionally, since the new `@highlighter` instance variable is only used in one place where it gets interpolated into a string, I have changed the instance variable to do the interpolation only once, saving additional string operations.

ashmaroli:
> have changed the instance variable to do the interpolation only once

That's even better!! :sparkles:

PR #650, opened 2020-04-10
Replace code snippet in "Using KaTeX instead of MathJax" section with a non-jQuery equivalent

doersino:
This isn't the fastest or most elegant solution, but it sure beats loading jQuery just for this purpose.


gettalong:
Thank you for this change. It would have been useful but the way the mathjax math engine outputs equations will change with the next version and therefore this text will also change accordingly.

PR #665, opened 2020-05-14
Return empty string if argument is an empty hash

ashmaroli:
When `attr == {}`, then end result to `Kramdown::Utils::Html#html_attributes(attr)` is always `""`, but involves allocation of intermediate empty array and the returned new empty string.

These allocations can be eliminated by returning a single frozen string every time `attr` is an empty hash.


gettalong:
Thanks!

Issue #48, opened 2013-04-18
figure element


iNerdier:
I’ve been trying to write lately with kramdown's markdown parser and I’ve come across two problems that seem to unsolvable in the current release.

One is that there seems to be no way to stop kramdown from wrapping image links in <p> tags. This is often fine but when paired with something like

``` css
p + p {
    text-indent: 1em;
}
```

the results are less than happy, images inherit unwanted text stylings and go all over the shop.

Markdown is smart enough not to add tags to existing block level elements, could the same method be used to allow specifying of stand-alone images?

If this is unfeasible or undesirable, could it be possible to allow it via an option ( say :image_block => true )

The second problem is attempting to use html5's new <figure> and <figcaption> elements. According to [can I use it](http://caniuse.com/#feat=html5semantic) all major browsers now support them and they’re certainly a great help in writing articles with images that need citations or have direct references to text.

Looking through google to see if anyone has attempted it I found https://github.com/michelf/php-markdown/wiki/HTML5-update which seems to have a pretty useful starting point for working out a syntax to use them.

Would either of these problems be something we might be able to do something about?

9peppe:
I'm probably late to this, but I like *a lot* how Pandoc handles this: if an image is in a paragraph on its own, it becomes a figure, and its alt-text is used as caption.

RoyiAvital:
Any chance having syntax for Figure?

Something like using the `alt-text` as caption as suggested by @9peppe ?

Thank You.

gettalong:
I can imagine providing a new option for changing the behaviour of how paragraphs with only an image are converted, similar to how the LaTeX converter currently does it - see https://github.com/gettalong/kramdown/blob/master/lib/kramdown/converter/latex.rb#L71

@9peppe / @RoyiAvital What should the output for the following kramdown fragment should be if the option to convert standalone images is activated?

~~~ markdown
This is a para with an ![image](some.jpg).

![standalone image](some.jpg)
~~~

Note that the alternative text of the image *cannot* contain markup, it is just text.

9peppe:
I'd say

    <p>This is a para with an <img alt="image" src="some.jpg" /></p>

    <figure id="someid">
      <img alt="standalone image" src="some.jpg" />
      <figcaption>standalone image</figcaption>
    </figure>

RoyiAvital:
3 Remarks (Though I'm not sure possible).

 1. Could we have a flag of some kind to whether use Figure or regular Image element?  
[...]

PR #667, opened 2020-05-20
Rendered Figure Captions

tyehle:
This builds on the changes discussed in issue #48 about wrapping images in a `figure` tag instead of a `p`. The current implementation works well and doesn't add a lot of extra complication, but with a few small changes I think it could be more flexible without being more complex.

## What I was trying to accomplish
- Don't make things more complicated
- Allow rendering in captions
- Allow alt text without a caption
- Keep the ability to put an `#id` or a `.class` on the `<figure>`

## Changes
There are three inputs that are different with these changes:

### A standalone image by itself
```markdown
![alt text](some.jpg){: standalone #id .class attr="value"}
```
The `alt text` is not taken as the image caption, instead there is no caption at all. The `#id` and `.class` stay on the `` instead of being copied to the `<figure>`. This is not a problem because a block IAL can fill this need.

### A standalone image with a block IAL
```markdown
![alt text](some.jpg){: standalone #id .class attr="value"}
{: #figure-id .figure-class figure-attr="value"}
```
The current implementation drops any attributes on the block instead of adding them to the figure. Implementing this seems like a nice solution to some of the concerns about id and class attributes on the image in #48.

### A standalone image with markdown after it
```markdown
![alt text](some.jpg){: standalone #id .class attr="value"}
Some [rendered](/link/location) image caption
{: .figure-class}
```
The current implementation does not make this a figure, but instead a paragraph with an inline image. Instead this would become a figure with any markup before the next blank line becoming a figure caption. This is similar to caption recommendations on [stack overflow](https://stackoverflow.com/a/30366422) and [Gitlab flavored markdown](https://about.gitlab.com/handbook/markdown-guide/#images) so it seems like people who have used such solutions would not find it all that surprising. It also allows rendering in the caption and allows the alt text to be different (another issue brought up in #48).

### DIFF
--- a/lib/kramdown/converter/html.rb
+++ b/lib/kramdown/converter/html.rb
@@ -86,24 +86,28 @@ module Kramdown
       def convert_p(el, indent)
         if el.options[:transparent]
           inner(el, indent)
-        elsif el.children.size == 1 && el.children.first.type == :img &&
+        elsif el.children.first.type == :img &&
             el.children.first.options[:ial]&.[](:refs)&.include?('standalone')
-          convert_standalone_image(el.children.first, indent)
+          convert_standalone_image(el, indent)
         else
           format_as_block_html("p", el.attr, inner(el, indent), indent)
         end
       end
 
-      # Helper method used by +convert_p+ to convert a paragraph that only contains a single :img
+      # Helper method used by +convert_p+ to convert a paragraph that starts with a standalone :img
       # element.
       def convert_standalone_image(el, indent)
-        attr = el.attr.dup
-        figure_attr = {}
-        figure_attr['class'] = attr.delete('class') if attr.key?('class')
-        figure_attr['id'] = attr.delete('id') if attr.key?('id')
-        body = "#{' ' * (indent + @indent)}<img#{html_attributes(attr)} />\n" \
-          "#{' ' * (indent + @indent)}<figcaption>#{attr['alt']}</figcaption>\n"
-        format_as_indented_block_html("figure", figure_attr, body, indent)
+        first, *rest = el.children
+
+        body = +"#{' ' * (indent + @indent)}<img#{html_attributes(first.attr)} />\n"
+
+        if rest.size > 0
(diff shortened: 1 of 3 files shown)


gettalong:
Thanks you for the pull request! However, it is not so easy:

* A standalone image by itself -> This would change existing behaviour without being a bug fix which is a no-go in most cases. 

* A standalone image with a block IAL -> Good idea, I would accept a pull request implementing this functionality. If an ID/class is set on the image and on the block, the ID/class from the image should not be propagated to the `<figure>` tag.

* A standalone image with markdown after it -> This is not a "standalone image" anymore but one embedded in a paragraph. Changing the semantics of this is definitely a no-go.

tyehle:
I see what you mean about semantics, but I didn't find any mention of standalone images on https://kramdown.gettalong.org/syntax.html. Am I looking in the wrong place?

Seems like a better way to go about this is by adding new behavior rather than changing existing behavior. Maybe a new block level `{:figure}` annotation instead of changing the way the inline `{:standalone}` annotation works?

gettalong:
Since standalone images are only an interpretation of a paragraph with an image, the only apply to converters, not the parser itself. So have a look at https://kramdown.gettalong.org/converter/html.html#standalone-images.

Yes, a `{:figure}` annotation would theoretically work but I wouldn't implement it in kramdown core. The reason is: Why stop at figure? Want a section by (ab)using a blockquote with the `{:section}` block IAL?

Having the standalone image is already a bit problematic from the semantic view of things. So I would rather not add other things like it...

tyehle:
That makes sense. Is there a way to implement this outside of the kramdown core? I saw some documentation about extensions; is this what you are referring to?

tyehle:
I created a different PR for the bug fix when there is a block IAL #668

gettalong:
You can subclass the HTML converter class and create a new one based on it. Then you just use this new converter instead of the base HTML converter.

Accepted #668, so I will close this.

PR #668, opened 2020-05-20
Block IAL for Standalone Images

tyehle:
Only the good parts from #667.

This changes the behavior of a standalone image with a block IAL:
``` markdown
![alt text](some.jpg){: standalone #id .class attr="value"}
{: #figure-id .figure-class figure-attr="value"}
```
The current implementation drops any attributes on the block and copies `#id` and `.class` attributes from the image to the `<figure>`.

With this change if there is an `#id` or `.class` on the block the annotations on the image remain with the image, and all block annotations are always copied to the `<figure>` instead of being deleted.


gettalong:
@tyehle Yes, thank you - I will have a look at it when I have some time!

gettalong:
@tyehle Thank you for your contribution. This will be in the next release.

PR #673, opened 2020-06-29
Update setup.rb

danielbair:
setup.rb:283: warning: key "bin-dir" is duplicated and overwritten on line 284


gettalong:
Thanks for your contribution - I have cherry-picked your commit for the next release.

PR #678, opened 2020-07-20
Document EOB marker

Convincible:
I spent hours trying to understand why this caret appeared here. Please can this be better documented? – ideally right here in the code


gettalong:
It is documented in the syntax documentation, see https://kramdown.gettalong.org/syntax.html#eob-marker. Putting the comment where you placed it is arbitrary since this is used in other places, too.

Convincible:
Fair enough. Can I ask if you're generally against adding more comments to the codebase to help others understand what certain blocks do?

In this case, rather than doubling the documentation for `^` arbitrarily, it would be useful instead to perhaps have a comment against the whole conditional like: "Check whether two blocks that shouldn't merge are next to each other and so need an EOB"

gettalong:
No, I'm not against adding comments where useful. The latter case you mention would be okay.

PR #679, opened 2020-07-20
Add parent info to element opts + nested list bugfix

Convincible:
Element opts hash carrying info about the parent is generally useful.

Immediate use case is simple fix for a bug where nested lists don't realise they are nested, and hence newlines are added after the end of a nested list (within the main list) when this is of course unnecessary and can cause rendering issues.


gettalong:
All elements are pushed onto the `@stack` array, so adding that extra information is not necessary. You can just use `@stack.last.type` as seen on the line directly above your change (https://github.com/gettalong/kramdown/pull/679/files#diff-fd924b970620e60748d914e517ebeed4L45).

Convincible:
Aha! Thanks. I overlooked the purpose of `@stack`.

Convincible:
New PR for the other part of this PR – #680

PR #680, opened 2020-07-21
Don't add newline after a nested list

Convincible:
Checks whether element is a list, and if that list is nested within another list (i.e. the list's parent is a list item). In this case we should not add another newline.

The change results in nested lists appearing as:

```
* First item
  * Sub item 1
    * Sub sub item
  * Sub item 2
* Second item
  * Sub item
```

rather than:

```
* First item
  * Sub item 1
    * Sub sub item

  * Sub item 2

* Second item
  * Sub item
```

### DIFF
--- a/lib/kramdown/converter/kramdown.rb
+++ b/lib/kramdown/converter/kramdown.rb
@@ -43,7 +43,8 @@ module Kramdown
         elsif el.block? &&
             ![:li, :dd, :dt, :td, :th, :tr, :thead, :tbody, :tfoot, :blank].include?(el.type) &&
             (el.type != :html_element || @stack.last.type != :html_element) &&
-            (el.type != :p || !el.options[:transparent])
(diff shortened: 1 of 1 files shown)


Convincible:
Now properly using the `@stack` variable

gettalong:
Thank you - I will have a more in-depth look later!

Convincible:
Has this bug been fixed @gettalong ?

gettalong:
@Convincible No, this change is not yet included but looks good! Please add some test and then it is good to go - thanks!

cabo:
> @Convincible No, this change is not yet included but looks good! Please add some test and then it is good to go - thanks!

Where would such a test go?
I'd really like to have this change.

gettalong:
Thanks! I have added a test case and this will be in the next release.

@cabo FYI Next release within a week.

PR #681, opened 2020-07-23
fix: fix LaTeX entity conversion group location

aschmitz:
This correctly places the `{}` group after the LaTeX commands used for conversion, rather than before. It appears this has been broken since edbd1fdf.

Also updates the span/text_substitutions/entities test to include `&lt;span&gt;`, which will be converted by the old converter to `{}\textlessspan{}\textgreater`, which should instead be `\textless{}span\textgreater{}`. The former will fail to compile, and should throw an error.

### DIFF
--- a/lib/kramdown/converter/latex.rb
+++ b/lib/kramdown/converter/latex.rb
@@ -517,7 +517,7 @@ module Kramdown
         8194 => ['\hskip .5em\relax'],
         8195 => ['\quad'],
       } # :nodoc:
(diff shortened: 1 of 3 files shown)


gettalong:
Thanks for the pull request! Just fyi: The commit you mentioned is not the one that introduced the bug but [this one is](https://github.com/gettalong/kramdown/commit/42280631fc748fae952e7bc60058cfbc2129a422#diff-1d8acc5d3b173b4a08fdc43990f6abf4L514-R520).

gettalong:
Thanks for your contribution, this will be in the next release.

PR #682, opened 2020-07-23
Bugfix: th and td elements can be empty

Convincible:
It is possible and valid to have an empty `<td></td>` element (similarly with `<th>`) and would be invalid to close it as `<td />`.

This change adds `th` and `td` to the list of elements that must be explicitly closed even if they have no content.

It avoids a bug where the invalid output `<td />` will be generated if a table cell has no content.

### DIFF
--- a/lib/kramdown/converter/kramdown.rb
+++ b/lib/kramdown/converter/kramdown.rb
@@ -182,7 +182,7 @@ module Kramdown
         result << inner(el, opts) << "\n"
       end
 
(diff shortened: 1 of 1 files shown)


gettalong:
Thanks for your contribution! Could you please add a test case in `test/testcases/block/14_table`?

gettalong:
Thanks for opening this issue - I have fixed this for the next release.

Issue #692, opened 2020-11-05
RFC: Option to make footnotes and footnote backlinks accessible


stevenjmesser:
Documents published on GOV.UK are authored in Markdown, and it was recently discovered that footnotes and footnote backlinks are not handled well by screen readers used by user with accessible needs.

To fix this, the Accessibility team on GOV.UK changed the way footnotes are handled. There's more details in these pull requests: https://github.com/alphagov/govspeak/pull/192 and https://github.com/alphagov/govspeak/pull/198

I'm raising this issue to see how others feel about adopting a similar change into Kramdown (or making it optional) in order to make thousands of blogs and websites more accessible to people using screen readers.

Thanks in advance for your comments!

stevenjmesser:
I've raised https://github.com/gettalong/kramdown/pull/693.

gettalong:
Thanks for the issue and the pull request!

I get what you want to accomplish with the issue and pull request. If the resulting HTML still validates, it would be okay for me.

However, I can't accept the current implementation since it is language dependent, i.e. the contents of the label are in English which is not suitable in a general case. If there is a language agnostic solution, I will have a more in-depth look at everything.

BryceStevenWilley:
I would also love something like this, have been trying to make my blog more accessible. I'm not sure there's a good language agnostic solution however, as the main issue is the fact that numbers and backlinks don't have enough information for context.

The linked issues are specifically changing the footnote link text, not adding an aria-label. IMO that's a better accessibility fix, as it also helps sighted users. The ability to do this exists in kramdown now: I was able to change the back links to match the linked issue by using the `footnote_backlink` option (added this to my jekyll `_config.yml`:

```yml
kramdown:
  footnote_backlink: "[go to where this is referenced]"
```

However, there's not an equivalent option for changing the link text. `footnote_prefix` seems like it should, but it only adds to the id of the elements. I might try to extend #693 to add another option, `footnote_link_text`, to be text that's added in front of the number in the footnote (maybe an `sprint`-able string, so you could do things like `[footnote 1]`), and open a PR soon.

stevenjmesser:
Seemingly addressed in https://github.com/gettalong/kramdown/pull/773

PR #693, opened 2020-11-07
Add aria-labels to footnotes and footnote backlinks

stevenjmesser:
Related to https://github.com/gettalong/kramdown/issues/692, here's a pull request to make footnotes accessible to screen readers.

### DIFF
--- a/lib/kramdown/converter/html.rb
+++ b/lib/kramdown/converter/html.rb
@@ -302,7 +302,7 @@ module Kramdown
           @footnotes_by_name[name] = @footnotes.last
         end
         "<sup id=\"fnref:#{name}#{repeat}\" role=\"doc-noteref\">" \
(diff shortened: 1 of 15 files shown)


gettalong:
You are using `rel="footnote"` (lowercase) but in some tests "Footnote" is written. So this would need to be fixed.

Furthermore, I don't want to introduce an English language default which would mix languages when the document is not written in English. Is aria-label meant to be an "internal" reference like with the `role` attribute? Or is this something a screen reader would actual read out loud?

If it is the latter, we would need to introduce a new option like `footnote_labels` that defines the text of the labels, with the default being off.

kinoute:
any update on this?

stevenjmesser:
I need to think about how to address @gettalong's comments as I'm not a developer, @kinoute! But do you have any ideas? I think I'd need to think about multilingual support most...

PR #703, opened 2021-02-17
Add support for converting more HTML elements

nikolai-b:
Allow sub, sup and u to be treated in the same way as em and strong

### DIFF
--- a/lib/kramdown/converter/html.rb
+++ b/lib/kramdown/converter/html.rb
@@ -314,10 +314,13 @@ module Kramdown
         end
       end
 
-      def convert_em(el, indent)
+      def convert_decoration_generic(el, indent)
         format_as_span_html(el.type, el.attr, inner(el, indent))
       end
-      alias convert_strong convert_em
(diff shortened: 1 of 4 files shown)


gettalong:
Can you explain what you what to achieve? Because

~~~
$ kramdown -o kramdown -i html
<p>This is <sub>sub</sub> and this is <sup>sup</sup>,<u>
unarticulated</u> or not it works a treat!</p>
^d
This is <sub>sub</sub> and this is <sup>sup</sup>,<u> unarticulated</u>
or not it works a treat!
~~~

already works.

nikolai-b:
Internally the `<sub>` et al. are all being represented as HTML elements, not as native kramdown elements.  I want to export them to PDF so would like to convert them all to native elements.

gettalong:
The internal element types are fixed based on what the kramdown syntax offers. Since there are no sub/sup/u elements in kramdown, no such elements should be created.

If you need to convert such HTML elements, override `convert_html_element` and do something special with those elements.

PR #705, opened 2021-03-10
Handle the Kramdown::Document.new input option like other options

jamiecobbett:
I was trying to understand how to switch parsers, but I was confused when I
couldn't find `input` in the Options documentation:
https://kramdown.gettalong.org/options.html I was instead able to find it in
the kramdown-gfm documentation: https://github.com/kramdown/parser-gfm but I
still wanted to know what the default value was - and to check that it was a
supported part of the API.

It was only after looking at the source code for a while that I realised
it is documented under `Kramdown::Document.new` as a special case.

Aside from helping people with the problems I encountered above, this means
this argument is parsed and validated in a more standard way.

### DIFF
--- a/lib/kramdown/document.rb
+++ b/lib/kramdown/document.rb
@@ -86,16 +86,11 @@ module Kramdown
     # Create a new Kramdown document from the string +source+ and use the provided +options+. The
     # options that can be used are defined in the Options module.
     #
-    # The special options key :input can be used to select the parser that should parse the
-    # +source+. It has to be the name of a class in the Kramdown::Parser module. For example, to
-    # select the kramdown parser, one would set the :input key to +Kramdown+. If this key is not
-    # set, it defaults to +Kramdown+.
-    #
     # The +source+ is immediately parsed by the selected parser so that the root element is
     # immediately available and the output can be generated.
     def initialize(source, options = {})
       @options = Options.merge(options).freeze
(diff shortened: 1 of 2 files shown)


gettalong:
The parser classes and the converter classes are not part of the options but part of the API interface. You create a `Kramdown::Document` and provide the `:input` argument for parsing, then use one of the `to_???` methods for conversion.

However, the `Kramdown::Document` class is just a convenience API to make the standard use case very simple. You can nearly as easily use the parser and converter classes directly, by providing the source string and the options to the parser, then the created root element and options to the converter.

As you can see creating a new option for the input just doesn't make sense.

jamiecobbett:
I guess we'll have to disagree on this one. In my use case, I am to continuing to call the same `Kramdown::Document.new`, but dynamically switching between parsers - so calling the lower-level parsers isn't relevant to my context. 

I also note that it does seem to work when passed in as an option to the binary, for example:
`kramdown --input=GFM README.md`

Where is this/should this be documented?

gettalong:
Selecting a different parser or converter with the `kramdown` binary works as expected. If you need to use a parser/converter that is not part of the kramdown distribution, you need to use the `-x` option to activate the extension. So in case of GFM you have to do `kramdown -x parser-gfm -i GFM README.md`.

PR #708, opened 2021-03-14
Restrict Rouge formatters to Rouge::Formatters namespace

stanhu:
ff0218aefcf00cd5a389e17e075d36cd46d011e2 added support for specifying custom Rouge formatters with the constraint that the formatter be in the`Rouge::Formatters` namespace, but it did not actually enforce this constraint. For example, this is valid:

```ruby
Rouge::Formatters.const_get('CSV')
=> CSV
```

Adding the `false` parameter to `const_get` prevents this:

```ruby
Rouge::Formatters.const_get('CSV', false)
NameError: uninitialized constant Rouge::Formatters::CSV
```

### DIFF
--- a/lib/kramdown/converter/syntax_highlighter/rouge.rb
+++ b/lib/kramdown/converter/syntax_highlighter/rouge.rb
@@ -70,7 +70,7 @@ module Kramdown::Converter::SyntaxHighlighter
       when Class
         formatter
       when /\A[[:upper:]][[:alnum:]_]*\z/
-        ::Rouge::Formatters.const_get(formatter)
+        ::Rouge::Formatters.const_get(formatter, false)
       else
         # Available in Rouge 2.0 or later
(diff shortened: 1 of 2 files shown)


gettalong:
Thanks! I think this makes sense also from a security point of view.

stanhu:
@gettalong Thanks! Would you mind merging and tagging a new release?

gettalong:
There will be a release in due time.

gettalong:
Thank you - merged and will be in the upcoming release.

aleksandrs-ledovskis:
Source of CVSS 9.9 [in GitLab](https://gitlab.com/gitlab-org/gitlab/-/commit/179329b5c3c118924fb242dc449d06b4ed6ccb66). 

I see no reason to keep it disconnected/under wraps once the patch (and consequently the vector) is known.

rfrohl:
I requested a CVE for this, will post an update once I got one assigned.

PR #726, opened 2021-09-18
Allow the text surrounded by square brackets with IAL to convert to the span element.

akinori-ichigo:
I made it to convert the text surrounded in square brackets with IAL to a plain span element.
For example,  "This is [CLASS_A]{:.class-a} text. This is [NO IAL] text." is converted to "\<p\>This is \<span class="class-a"\>CLASS_A\</span\> text. This is [NO IAL] text.\</p\>".
I think this feature get kramdown more expressive and useful.

### DIFF
--- a/lib/kramdown/converter/html.rb
+++ b/lib/kramdown/converter/html.rb
@@ -277,6 +277,14 @@ module Kramdown
         "<img#{html_attributes(el.attr)} />"
       end
 
+      def convert_span(el, indent)
+        if el.attr.empty?
+          "[#{inner(el, indent)}]"
+        else
+          format_as_span_html('span', el.attr, inner(el, indent))
+        end
+      end
+
       def convert_codespan(el, _indent)
         attr = el.attr.dup
         lang = extract_code_language(attr)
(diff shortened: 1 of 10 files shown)


gettalong:
This conflicts with current syntax:

~~~
This is [Class_a]{:class-a} test other.

[Class_a]: link.html
~~~

Output:

~~~
<p>This is <a href="link.html">Class_a</a> test other.</p>
~~~

The IAL is not applied to the link which should probably be done but besides that changing the syntax is not possible.

akinori-ichigo:
I think the IAL has been applied to the link in Kramdown 2.3.1:

```
$ gem list kramdown

*** LOCAL GEMS ***

kramdown (2.3.1)

$ pry
[1] pry(main)> require "kramdown"
=> true
[2] pry(main)> Kramdown::Document.new(<<EOS).to_html
[2] pry(main)* This is [Class_a]{:.class-a} test other.
[2] pry(main)* 
[2] pry(main)* [Class_a]: link.html
[2] pry(main)* EOS
=> "<p>This is <a href=\"link.html\" class=\"class-a\">Class_a</a> test other.</p>\n\n"
[3] pry(main)> 
```

gettalong:
@cabo No, there is no such markup in kramdown. You would need to define a custom markup for this.

Issue #663, opened 2020-05-04
Shared mutable state is not thread safe.


ioquatix:
https://github.com/gettalong/kramdown/blob/4458c23153884cb576e4a57739e191317256abea/lib/kramdown/converter/html.rb#L55

gettalong:
I'm not that well versed with multi-threading, but can it be a problem in this case? The worst case is that the key is set twice (multiple times) to the same value, isn't it?

ioquatix:
I did a live stream about this: https://www.youtube.com/watch?v=Wu9LRNOc5pQ

Yes, it can be a problem.

I would suggest:

1. Benchmark the use of a cache or not. Is it providing a statistically significant improvement?
2. Pre-generate the hash table and freeze it if the cache is valuable.

gettalong:
Thanks for the live stream - you make it very clear :+1:

There is significant difference in performance, probably mostly due to less created objects and less garbage collection.

Here is a YAML file for benchmark-driver:

~~~yaml
prelude: |
  hash = {:root=>1, :hr=>1, :p=>315, :text=>631, :typographic_sym=>21, :html_element=>1, :blank=>387, :header=>45, :strong=>5, :smart_quote=>24, :ul=>21, :li=>74, :a=>59, :em=>15, :dl=>4, :dt=>14, :dd=>14, :codespan=>88, :codeblock=>94, :blockquote=>29, :math=>1}
  dispatcher = Hash.new {|h, k| h[k] = "convert_#{k}" }
  mutex = Mutex.new
  mutex_dispatcher = Hash.new {|h, k| mutex.synchronize { h[k] = "convert_#{k}" } }


benchmark:
  direct: |
    hash.each {|name, calls| calls.times { "convert_#{name}" } }

  hash: |
    hash.each {|name, calls| calls.times { dispatcher[name] } }

  hash with mutex: |
    hash.each {|name, calls| calls.times { mutex_dispatcher[name] } }
~~~

Here is the result on Ruby 2.6.5:

~~~
Warming up --------------------------------------
              direct     3.569k i/s -      3.894k times in 1.090981s (280.17μs/i)
                hash    12.532k i/s -     13.772k times in 1.098990s (79.80μs/i)
     hash with mutex    12.308k i/s -     12.710k times in 1.032700s (81.25μs/i)
Calculating -------------------------------------
              direct     3.523k i/s -     10.707k times in 3.039080s (283.84μs/i)
[...]

PR #727, opened 2021-09-19
Allocate a single dispatcher for `Html` Converters

ashmaroli:
Mostly because I do not see the need to initialize a new Hash for every `Converter::Html` instance.

(kept the instance variable intact for backwards compatibility with extensions until next major version).

### DIFF
--- a/lib/kramdown/converter/html.rb
+++ b/lib/kramdown/converter/html.rb
@@ -32,6 +32,9 @@ module Kramdown
       include ::Kramdown::Utils::Html
       include ::Kramdown::Parser::Html::Constants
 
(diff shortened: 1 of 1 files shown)


gettalong:
See https://github.com/gettalong/kramdown/issues/663 why this won't be merged.

ashmaroli:
> .. why this won't be merged.

My bad! I should have checked the *blame* for the reason behind the implementation. Lesson learnt! ;)
On that note, did you benchmark using a *frozen dispatcher pre-filled with all the `:converter_*` keys?*

gettalong:
No problem :smile: I probably wouldn't have done the blame either.

I did only benchmark what's listed in that issue. And I wouldn't freeze the dispatcher because that may be a problem if someone extends the element set.

Issue #737, opened 2021-10-30
It is not clear how to write a multipoint entity in your entity list


StoneCypher:
Some HTML entities, such as `nsubE`, are represented as multiple unicode characters (in this case `U+2AC5` `U+0338`.)  This is particularly common in math symbols using the slash to strike through symbols.

It is not immediately clear to me how to represent that in [the kramdown entity list](https://github.com/gettalong/kramdown/blob/master/lib/kramdown/utils/entities.rb).

If you could tell me how to represent that one case please, I would happily extend it to the remainder.

gettalong:
The conversion from codepoint to character string is done like this:

~~~
[code_point].pack('U*')
~~~

This will create the correct string representation for any Unicode codepoint. So as long as the entity consists of a single code point, this will work.

Does that clear it up?

StoneCypher:
I apologize.  That isn't what I meant.

```
      ENTITY_TABLE = [
        [913, 'Alpha'],
        [914, 'Beta'],
        [915, 'Gamma'],
```

...

```
        [213, 'Otilde'],
        [214, 'Ouml'],
        [215, 'times'],
```

Please pretend for a moment that there was no dedicated capital-O umlaut character.  There is, of course; it's `U+00D6`, represented here as decimal 214.  But let's pretend there wasn't.

In Unicode, there is a dedicated combining diaresis, and you can attach it to other characters to construct the character you need.  As such, you could make the character with capital O `U+004F` then combining diaresis `U+0308`.

So for a moment, pretend please that I want to rewrite your Ouml rule to emit two codepoints.  In this case it's silly, but this is legitimately how quite a few entities (particularly in math) are written.  By example, `&nsubE;` - `Not subset-equal` - is written as `U+2288`, the dedicated math symbol, but really should be written as `U+10949` subset equal `U+338` negating slash (the logic symbol) instead.

And that's hard to think about, so we're lying, and talking about O umlaut.

If for some stupid reason I wanted to emit `U+004F` `U+0308` for `Ouml` in this table, how would I do it?

gettalong:
I see. This is not possible with how the entities are implemented in kramdown though it is easily doable by just doing `[code_point1, code_point2].pack('U*')`.

As far as I can see, however, all the HTML5 entities are just single-codepoint entities? So this should not be a problem here.

StoneCypher:
> 

There are a few.

| Name | Symbol | Codepoint |
| --- | --- | --- |
[...]

Issue #736, opened 2021-10-30
All HTML5 entities are missing


StoneCypher:
Edit: TL, DR?  [Patch in PRs](https://github.com/gettalong/kramdown/pull/738).

----

Hello.  This was reported in 2017, and it seems that a misunderstanding led to it not being fixed.

I attempted to write `&frac13;` in Github pages, and it did not go well.  Works on GH, but not on the rendered page.

![image](https://user-images.githubusercontent.com/77482/139557322-ff196803-1a1d-42cc-b7d2-90cb6d28a761.png)

Snooping led to Jekyll, and then here.  [And then I found this](https://github.com/gettalong/kramdown/issues/114#issuecomment-328128093).

1. Whether HTML is defined by the W3C or WhatWG has been unclear for about ten years.  It's probably actually WhatWG.
2. The [WhatWG entity list](https://html.spec.whatwg.org/multipage/named-characters.html) and the [W3C entity list](https://dev.w3.org/html5/html-author/charref) currently match.
3. When the WhatWG adds characters, the W3C adds them within a year.  There is no official stance here, but it has never not gone this way.
4. The W3C last added characters without prompting in 1997.
5. WhatWG is made of up W3C people.

The [Wikipedia list](https://en.wikipedia.org/wiki/List_of_XML_and_HTML_character_entity_references) is much more convenient than the two standards lists, because it's a sortable table which includes what standard the character was added during.  You can sort by that column to yield a list of what's missing.

When you do so, you get this:

| Name | Char | Codepoint | Source |
| --- | --- | --- | --- |
| Lang | ⟪ | U+27EA (10218) | HTML 5.0 |
| Rang | ⟫ | U+27EB (10219) | HTML 5.0 |
| loang | ⟬ |  U+27EC (10220) | HTML 5.0 |
| roang | ⟭ |  U+27ED (10221) | HTML 5.0 |
| xlarr, longleftarrow, LongLeftArrow | ⟵ |  U+27F5 (10229) | HTML 5.0 |
| xrarr, longrightarrow, LongRightArrow | ⟶ |  U+27F6 (10230) | HTML 5.0 |
| xharr, longleftrightarrow, LongLeftRightArrow | ⟷ |  U+27F7 (10231) | HTML 5.0 |
| xlArr, Longleftarrow, DoubleLongLeftArrow | ⟸ |  U+27F8 (10232) | HTML 5.0 |
[...]

PR #738, opened 2021-10-30
Shaky attempt to add the HTML5 entities

StoneCypher:
Hello.  I don't speak Ruby, and I don't know how to run the tests on this.  Use stink-eye heavily please.

I wrote this patch to attempt to address #736 , that all HTML5 entities are missing.  I want `&frac13;`.

Two assumptions are made in this patch which I am concerned about, above and beyond that I should be allowed anywhere near a keyboard.

1. Extended combined entities may be represented in their long form. This is separately addressed as #737.  I currently represent `&nsubE;` as `10949` which is acceptable, but it should be `10949` `338` instead.
2. Integer placements may be repeated.  Some entities have multiple names, such as `10878`, which can be called `&ngeqslant;` or `&nges;`.  I'm assuming I can just list them both, but I don't ruby, so I'm not 100% sure I'm reading that hash table's uniqueness criterion correctly.

Thanks kindly.


StoneCypher:
If you're curious, the method for producing this patch was:

1. Using `firefox`, control-select to pull the appropriate columns out of [this table](https://en.wikipedia.org/wiki/List_of_XML_and_HTML_character_entity_references), after sorting on `standard`
2. Postprocess with the below script in a browser console, because lazy

```javascript
const chars = `Lang  ⟪   U+27EA (10218)  HTML 5.0
Rang  ⟫   U+27EB (10219)  HTML 5.0
... rest of Firefox paste here ...
varsupsetneqq, vsupnE   ⫌︀  U+2ACC (10956), U+FE00 (65024)  HTML 5.0
nparsl  ⫽⃥  U+2AFD (11005), U+20E5 (8421)   HTML 5.0`
  .split('\n')
  .map(row => {
    const group = row.replace('   ', '  ');
    return group.split('  ')
  });



const makeRow = (name, num) => {
  const names = name.split(', ');
  return names.map(nm => `        [${num}, '${nm}'],`).join('\n');
}




console.log(
  chars.map(
    row => makeRow(row[0], parseInt(row[2].split('(')[1].split(')')[0]))
  )
    .join('\n')
);
```

StoneCypher:
@gettalong

gettalong:
@StoneCypher I will include this in the next released, though the expanded version that also handles multi-codepoint entities will have to wait.

StoneCypher:
OK :)

StoneCypher:
Thank you

PR #740, opened 2021-12-09
Make initial list label available for processor

cabo:
Making the initial list label available to the processor enables implementing behaviors such as deriving a starting number from the initial label.
(This should not impact any existing code.)

### DIFF
--- a/lib/kramdown/parser/kramdown/list.rb
+++ b/lib/kramdown/parser/kramdown/list.rb
@@ -69,6 +69,7 @@ module Kramdown
             eob_found = true
             break
           elsif @src.scan(list_start_re)
+            list.options[:first_list_marker] ||= @src[1].strip
             item = Element.new(:li, nil, nil, location: start_line_number)
             item.value, indentation, content_re, lazy_re, indent_re =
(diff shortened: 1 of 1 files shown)


gettalong:
Thanks!

* Yes, this will not impact any code or the output, so :+1: 

* As for semantics: Would it be better to call it `item_label` and assign it to all items? Then the list label can be determined by looking at the first item label.

cabo:
I specifically need the first label of the list (so I put it on the list).  
But putting a label on each item would only insignificantly increase the amount of code needed.

gettalong:
Hmm... okay, then we will only use the first one and if need be add the change for the others.

One thing to change, though: Could you name it `:first_list_marker`? In the syntax documentation it is always called "list marker" and since it is the first one and not something else, it should also have "first" somewhere in its name.

cabo:
I pushed an update with `:first_list_marker`.

gettalong:
Thanks for the changes - I will do a release for this in the near future.

gettalong:
Thanks - I have cherry-picked the commit!

PR #748, opened 2022-03-12
Make codespan delimiter available for processor

cabo:
Similar to PR #740:

Making the codespan delimiter available to the processor enables
implementing behaviors deriving variant behavior from using different
numbers of delimiters.
(This should not impact any existing code.)


gettalong:
Thanks - I have merged this!

cabo:
Thanks!  Looking forward to an eventual release with this and #740 ...

gettalong:
@cabo I just released a new version with the two enhancements.

PR #761, opened 2022-08-25
pass parser options into convert

ghepting:
When using the `.to_kramdown` converter, we were unable to pass the `:raw_text` option through to avoid escaping text such as `{foo_bar}` etc. This change passes all parser options through to the converter – not sure if we want that or another interface to pass the options through on `.to_kramdown` but this seemed okay with my basic testing. Looking for input/insight from maintainer to verify strategy here.

### DIFF
--- a/lib/kramdown/converter/base.rb
+++ b/lib/kramdown/converter/base.rb
@@ -104,7 +104,7 @@ module Kramdown
         if !converter.options[:template].empty? && converter.apply_template_before?
           apply_template(converter, '')
         end
-        result = converter.convert(tree)
+        result = converter.convert(tree, converter.options)
         if result.respond_to?(:encode!) && result.encoding != Encoding::BINARY
           result.encode!(tree.options[:encoding] ||
                          (raise ::Kramdown::Error, "Missing encoding option on root element"))


gettalong:
I'm not quite sure what you want to achieve by passing the options of the converter back to the converter as an argument to one of its methods:

~~~
converter.convert(tree, converter.options)
~~~

Could you provide a full example that doesn't work as you would expect?

ghepting:
Yeah I'll post reproduction scenario in the morning tomorrow absolutely

ghepting:
Okay so here's the before & after:
Screen Shot 2022-08-26 at 9 06 20 AM
Screen Shot 2022-08-26 at 9 10 14 AM

As you can see, I'm needing to get `raw_text: true` option passed all the way down to the converter.convert method in order to leverage the logic here: https://github.com/gettalong/kramdown/blob/master/lib/kramdown/converter/kramdown.rb#L77-L78 and not have this raw content get escaped. we're using some simple handlebars template tags within the plain text of our documents because it's only 1 phase of an ETL pipeline before some variables are injected for an email send.

Very open to input/education here if there's a better way to accomplish this that we've missed in our reverse engineering! :)

gettalong:
There is no `raw_text` option in kramdown, so it seems you are using a custom option. However, what you really try to use is a *converter specific and internal* option of the kramdown converter that has nothing to do with the general option mechanism.

What you wanna do is create a custom kramdown converter that handles your text nodes specially, i.e. by always outputting them as is. This, however, might lead to problems parsing the result as kramdown without a pre-processing step (which you probably have).

ghepting:
Ah, okay I'll just setup a custom converter that removes the escaped characters after the kramdown conversion so it's unobtrusive and a more clear additional processing step. Thanks for the time & guidance.

gettalong:
No problem. I think it is a clearer solution than (ab)using some internal state and makes it clearer what you want to achieve. Feel free to contact me if you run into problems with the converter.

PR #764, opened 2022-09-02
Adding the emphasis character to the options of the element.

ionactual:
This is a small change that will make it easier to support underlines when dealing with the document models.


gettalong:
Thanks @ionactual for the pull request - this is [merged](https://github.com/gettalong/kramdown/compare/master...devel).

PR #765, opened 2022-09-03
Add links to headers 🔗

mctaylorpants:
Hello! I've implemented a feature that I hope would be a good fit for kramdown: the ability to add links to headers.

### How it works
When the `header_links` option is set to `true` and `auto_ids` is also `true`, headers will receive an anchor tag - just like [the ones on a Github README page](https://github.com/gettalong/kramdown#readme-first)

```
## Example header
```

... becomes:

```
<h2 id="example-header"><a href="#example-header">Example header</a></h2>
```

Having a link in the header is handy for generating a permalink to share when there's no table of contents on the page.


mctaylorpants:
I've currently got two test failures in the suite - I can look into them further if the overall feature looks good:

```
TestFiles#test_/usr/src/app/test/testcases/block/04_header/with_header_links_text_to_html [/usr/src/app/test/test_files.rb:60]:
--- expected
+++ actual
@@ -16,7 +16,7 @@

 <h2 id=\"hallo-2\"><a href=\"#hallo-2\">hallO</a></h2>

-<h1>Header without ID</h1>
+<h1><a href=\"#\">Header without ID</a></h1>

 <h1 id=\"transliterated-day-la-vi-du\"><a href=\"#transliterated-day-la-vi-du\">Transliterated: Đây-là-ví-dụ</a></h1>

```

```
TestFiles#test_whether_Html_modifies_tree_with_file_/usr/src/app/test/testcases/block/04_header/with_header_links_text [/usr/src/app/test/test_files.rb:284]:
type mismatch.
Expected: :text
  Actual: :a
```

lib/kramdown/converter/html.rb
```diff
@@ -147,6 +147,14 @@ def convert_header(el, indent)
         if @options[:auto_ids] && !attr['id']
           attr['id'] = generate_id(el.options[:raw_text])
         end
+
+        if @options[:auto_ids] && @options[:header_links]
+          link = Element.new(:a, nil, nil)
+          link.attr['href'] = "##{attr['id']}"
+          link.children << el.children.pop
```
gettalong, line 154: A header element may have more than one children, e.g. `# This *is* something`.

lib/kramdown/converter/html.rb
```diff
@@ -147,6 +147,14 @@ def convert_header(el, indent)
         if @options[:auto_ids] && !attr['id']
           attr['id'] = generate_id(el.options[:raw_text])
         end
+
+        if @options[:auto_ids] && @options[:header_links]
```
gettalong, line 151: Option 'audo_ids' is not necessary for this feature to work, the header just needs to have an 'id' attribute set.

lib/kramdown/converter/html.rb
gettalong: And I would *not* make the whole header itself clickable since that would potentially interfere with a link in the header itself. Why not just generate an empty `<a>` tag at the beginning of the header and let the user style that to their liking? E.g. by showing some content when hovering over the header.

gettalong (commented):
Thanks for your pull request! This might indeed be something useful to have as a general kramdown feature.

mctaylorpants:
Thanks for the comments and suggestions, @gettalong! I've made some updates:

- Inserts an empty anchor tag as the first child in the header
- Removes the dependency on `auto_ids`, and just checks for the presence of an ID instead.

lib/kramdown/converter/html.rb
```diff
@@ -148,11 +148,10 @@ def convert_header(el, indent)
           attr['id'] = generate_id(el.options[:raw_text])
         end
 
-        if @options[:auto_ids] && @options[:header_links]
+        if @options[:header_links] && attr['id'].length > 0
```
gettalong, line 151: `attr['id'].length` will error out if there is no 'id' set. So better use something like `attr['id'].to_s.length > 0`.

gettalong (changes requested):
I think after that one change the PR is good to go! :+1:

gettalong:
@mctaylorpants Thanks! I have squashed the commits and the [changes](https://github.com/gettalong/kramdown/commit/da85e1e2e1f797f29f546f76d0ecc92270c6a0ec) will be in the next release.

CookiePLMonster:
Hello from 2025, just pointing something out - I just discovered this feature is now part of Kramdown, but it took me too long to realize that the version of Kramdown that I currently use (2.4.0) is too old for it, because the 2.5.0 release notes never mentioned it:
<https://kramdown.gettalong.org/news.html>

A new `footnote_link_text` option added in 2.5.0 was added, while this one was not. IMO I feel like it's big enough deal to consider retroactively adding it to that changelog.

PR #774, opened 2022-11-20
Add new option, `footnote_link_text`

### LINKED ISSUE #692: RFC: Option to make footnotes and footnote backlinks accessible

BryceStevenWilley:
`footnote_link_text` is a string that is directly inserted in front of the footnote number. It can also be a formatted string that the number is formatted into.

For example, if the string is "[footnote %s]", the footnote link will be "[footnote 1]".

It will determine it the option is a forat string by running it through `sprintf`; if the output is the same, then it's not a format string.

Also adds tests to make sure of the new output.

Fixes #692, and is a different attempt at #693.

### DIFF
--- a/lib/kramdown/converter/html.rb
+++ b/lib/kramdown/converter/html.rb
@@ -294,6 +294,7 @@ module Kramdown
       def convert_footnote(el, _indent)
         repeat = ''
         name = @options[:footnote_prefix] + el.options[:name]
+        link_text = @options[:footnote_link_text]
         if (footnote = @footnotes_by_name[name])
           number = footnote[2]
           repeat = ":#{footnote[3] += 1}"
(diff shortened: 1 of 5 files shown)


gettalong:
Thanks @BryceStevenWilley for the pull request!

What is the benefit of keeping the special case `formatted_link_text == link_text`? Wouldn't it be easier to omit that special case and make the default value `'%s'`.

BryceStevenWilley:
> What is the benefit of keeping the special case formatted_link_text == link_text?

It lets folks use `footnote_link_text` without using the printf formatting. If they don't use the formatting (when formatted link is the same as the link text), we'll add the number at the end, so the links will still read as distinct.

Making the default case `%s` is a good idea though! Makes the default behavior much clearer, and introduces people to the formatting feature. I can make that change soon, unless you want to make it yourself.

gettalong:
Thanks for the feedback! So I would go with the solution of `%s` being the default value. This also means the option description needs to be updated and there should be a validation block making sure that `%s` appears in the option value.

BryceStevenWilley:
> o I would go with the solution of %s being the default value. This also means the option description needs to be updated and there should be a validation block making sure that %s appears in the option value.

This is in now! Let me know if you want it changed at all, I tried to match the surrounding validation blocks, but wasn't certain how well I did.

gettalong:
Thanks for the pull request! This will be in the next release.

PR #781, opened 2023-03-13
Fix deprecated interface for YAML.safe_load

stanhu:
Passing in the permitted classes as the second argument was deprecated in Psych 3 (https://github.com/ruby/psych/pull/378) and removed in Psych 4 (https://github.com/ruby/psych/pull/487).

Ruby 3.1 ships with Psych 4, so to make the code work for Ruby 2.7 and up, use the `permitted_classes` keyword argument.

### DIFF
--- a/bin/kramdown
+++ b/bin/kramdown
@@ -111,7 +111,7 @@ end.parse!
 
 begin
   if config_file && File.exist?(config_file)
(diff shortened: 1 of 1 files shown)


stanhu:
Closing in favor of https://github.com/gettalong/kramdown/pull/779.

PR #807, opened 2024-06-22
fix: cast source to a string before checking encoding

roelbondoc:
Because of [this change](https://github.com/rails/rails/pull/51023) in rails, the template source that gets passed into here could be an `ActionView::OutputBuffer` which doesn't respond to `valid_encoding?`. The quick fix here is to cast it into a string first.


gettalong:
I'm sorry but the quick fix here is to follow the kramdown API documentation and provide a string to the method and not an object from some some other library:

https://github.com/gettalong/kramdown/blob/bd678ecb59f70778fdb3b08bdcd39e2ab7379b45/lib/kramdown/parser/base.rb#L89-L91

PR #808, opened 2024-06-27
Prevent unnecessary escaping of numbers

pwim:
When a line starts with a number that ends with a period, e.g. `1984.`, it needs to be escaped, as otherwise markdown interprets it as a list.

However, if a line begins with a number that includes a period in it, e.g. `1984.5`, it does not need to be escaped, as it won't be interpreted as a list.


gettalong:
Thanks! I removed the `\z` from the character class but otherwise used your pull request as is.

PR #810, opened 2024-07-18
[bugfix] Fix edge case in Kramdown converter

### LINKED ISSUE #809: [bug] Kramdown converter fails with NoMethodError: undefined method `type' for nil
Steps to reproduce:

```ruby
html = File.read("example.html")
doc, = Kramdown::Parser::Html.parse(html)
Kramdown::Converter::Kramdown.convert(doc)
```

will submit a PR with a fix

honzasterba:
weird combination of LI and P tags can cause the converter to crash

fixes #809


gettalong:
Thanks for the pull request! The fix was not fully correct so I implemented the fix and test case myself.

PR #819, opened 2024-12-11
Change `<summary>` to use span content model instead of block

otegami:
This PR proposes changing the content model for `<summary>` back to inline, ensuring cleaner output and a more intuitive user experience.

In HTML5, the `<summary>` element can be treated as either block or inline (phrasing) content. Currently, it's set as a block element, which can lead to formatting issues in general usage scenarios like the following.

## How to reproduce

Consider the following markdown file (summary.md).

```markdown

<details>
<summary>Hello</summary>
World
</details>
```

When processed with `bin/kramdown summary.md` with
the `--parse-block-html` option.

```console
$ bin/kramdown summary.md --parse-block-html
<h1 id="summary-test">Summary test</h1>

<details>
  <summary>
    <p>Hello&lt;/summary&gt;
World
&lt;/details&gt;</p>

  </summary>
</details>
Warning: Found no end tag for 'summary' (line 4) - auto-closing it
Warning: Found no end tag for 'details' (line 3) - auto-closing it
Warning: Found invalidly used HTML closing tag for 'summary' on line 4
Warning: Found invalidly used HTML closing tag for 'details' on line 6
```

### Expected output

```console
$ bin/kramdown summary.md --parse-block-html
<h1 id="summary-test">Summary test</h1>

<details>
  <summary>Hello</summary>
  <p>World</p>
</details>
```

## Suggestion

By treating `<summary>` as an inline span element rather than a block element, we can align with the expected output and common usage patterns. This change is compliant with the HTML5 specification, which allows `<summary>` to be treated as inline content.

### DIFF
--- a/lib/kramdown/parser/html.rb
+++ b/lib/kramdown/parser/html.rb
@@ -38,10 +38,10 @@ module Kramdown
         HTML_CONTENT_MODEL_BLOCK = %w[address applet article aside blockquote body
                                       dd details div dl fieldset figure figcaption
                                       footer form header hgroup iframe li main
-                                      map menu nav noscript object section summary td]
+                                      map menu nav noscript object section td]
         HTML_CONTENT_MODEL_SPAN  = %w[a abbr acronym b bdo big button cite caption del dfn dt em
                                       h1 h2 h3 h4 h5 h6 i ins label legend optgroup p q rb rbc
-                                      rp rt rtc ruby select small span strong sub sup th tt]
+                                      rp rt rtc ruby select small span strong sub summary sup th tt]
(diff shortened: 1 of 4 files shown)


gettalong:
Using `--parse-block-html` is not advised and should only be used in special circumstances. Anyway, it is better to use the `markdown="span"` attribute to handle such cases where the `<summary>` element should be treated as holding span level content:

~~~
<details>
<summary markdown="span">Hello</summary>
World
</details>
~~~

results in the following when run with `kramdown --parse-block-html`:

~~~ html
<details>
  <summary>Hello</summary>
  <p>World</p>
</details>
~~~

PR #824, opened 2025-01-20
Switch from Travis CI to GitHub CI

Krinkle:
It seems Travis CI is no longer available for pull requests and forks to run tests. In order for this to work by default and give feedback early, try GitHub CI instead.


Krinkle:
Example results: https://github.com/Krinkle/kramdown/actions/runs/12875286798

Krinkle:
Results: https://github.com/Krinkle/kramdown/actions/runs/12875590161

I removed Ruby 2.3 and 2.4 from the test matrix because those were failing with:

```
 kramdown was resolved to 2.5.1, which depends on
      Ruby (>= 2.5)

  Current Ruby version:
    Ruby (= 2.4.10)
```

PR #841, opened 2026-01-29
Fix 'character class has duplicated range' warning.

### LINKED ISSUE #821: Ruby warning: character class has duplicated range: `/[\s\p{Z}]+/`
Caught from third-party gem (`r18n-core`):

```
/home/alex/.rbenv/versions/3.3.6/lib/ruby/gems/3.3.0/gems/kramdown-2.5.1/lib/kramdown/parser/base.rb:60: warning: character class has duplicated range: /[\s\p{Z}]+/
```

Also I'd suggest to turn on Ruby warnings for test: https://github.com/gettalong/kramdown/commit/6c222cc7552807b517048630e15d32fc2f6a7b00

zenspider:
Fixes #821.

### DIFF
--- a/lib/kramdown/parser/base.rb
+++ b/lib/kramdown/parser/base.rb
@@ -57,7 +57,7 @@ module Kramdown
 
         @root.options[:abbrev_defs].default_proc = @root.options[:abbrev_attr].default_proc =
           lambda do |h, k|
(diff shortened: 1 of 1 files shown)


gettalong:
Thanks but the fixed regexp is not equivalent to the regexp from before as \s encompasses more than just the space character. I have fixed it with 5a41893.

